PR #18, opened 2020-02-26
Add test for SpatialPredicatePartitioner

errearanhas:
Tried to implement two saptial predicates: one for "inside radius check" and other for "inside rectangle check".

### DIFF
--- a/test/partitioning.jl
+++ b/test/partitioning.jl
@@ -197,7 +197,7 @@
 
   @testset "SpatialPredicatePartitioner" begin
     g = RegularGrid{Float64}(10,10)
-
+    
     # check if each partition has only 1 point
     sp = SpatialPredicatePartitioner((x,y) -> norm(x-y) < 1.0)
     s = subsets(partition(g, sp))
--- a/test/partitioning.jl
+++ b/test/partitioning.jl
@@ -197,19 +197,27 @@
 
   @testset "SpatialPredicatePartitioner" begin
     g = RegularGrid{Float64}(10,10)
-    
-    # check if each partition has only 1 point
+
+    # check if there are 100 partitions, each one having only 1 point
     sp = SpatialPredicatePartitioner((x,y) -> norm(x-y) < 1.0)
     s = subsets(partition(g, sp))
     @test length(s) == 100
-
-    # check if points x and y belong to the rectangle [0.,5.]x[0.,5.]
+    for d in partition(g, sp)
+      @test npoints(d) == 1
+    end
+    # defining a predicate to check if points x and y belong to the rectangle [0.,5.]x[0.,5.]
     pred(x, y) = all([0.,0.] .<= x .<=[5.,5.]) && all([0.,0.] .<= y .<= [5.,5.])
     sp = SpatialPredicatePartitioner(pred)
-    s = subsets(partition(g, sp))
+    p = partition(g, sp)
+    s = subsets(p)
+    n = npoints.(p)
 
+    # a square [0.,5.]x[0.,5.] in a 10x10 RegularGrid has to contain 36 points
+    # all other points have to be each in its own partition (64 partitions)
     @test length(s) == 65
     @test maximum(length.(s)) == 36
+    @test count(i->i==1, n) == 64
+    @test count(i->i==36, n) == 1
   end
 
   @testset "ProductPartitioner" begin
--- a/test/partitioning.jl
+++ b/test/partitioning.jl
@@ -196,7 +196,20 @@
   end
 
   @testset "SpatialPredicatePartitioner" begin
-    # TODO
+    g = RegularGrid{Float64}(10,10)
+
+    # check if each partition has only 1 point
+    sp = SpatialPredicatePartitioner((x,y) -> norm(x-y) < 1.0)
+    s = subsets(partition(g, sp))
+    @test length(s) == 100
+
+    # check if points x and y belong to the rectangle [0.,5.]x[0.,5.]
+    pred(x, y) = all([0.,0.] .<= x .<=[5.,5.]) && all([0.,0.] .<= y .<= [5.,5.])
+    sp = SpatialPredicatePartitioner(pred)
+    s = subsets(partition(g, sp))
+
+    @test length(s) == 65
+    @test maximum(length.(s)) == 36
   end
 
   @testset "ProductPartitioner" begin
--- a/test/partitioning.jl
+++ b/test/partitioning.jl
@@ -205,15 +205,16 @@
     for d in partition(g, sp)
       @test npoints(d) == 1
     end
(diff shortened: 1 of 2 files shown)


test/partitioning.jl
```diff
@@ -196,7 +196,31 @@
   end
 
   @testset "SpatialPredicatePartitioner" begin
-    # TODO
+    g = RegularGrid{Float64}(10,10)
+    R = 5.0
+    SPred1(x,y) = norm(x-y) < R
```
juliohm: This predicate is a small one-line function, please use a lambda to construct the partitioner:

```Julia
p = partition(g, SpatialPredicatePartitioner((x, y) -> norm(x-y) < 5.0)
```

And then take the subsets directly in one line:

```julia
s = subsets(partition(g, p))
```

test/partitioning.jl
```diff
@@ -196,7 +196,31 @@
...
+
+    p = partition(g, SpatialPredicatePartitioner(SPred1))
+    s = subsets(p)
+    testArray = sample(s)
```
juliohm: This test should be exhaustive for all sets in `s`. Please write a for loop instead of sampling a single set.

test/partitioning.jl
```diff
@@ -196,7 +196,31 @@
...
+    @test (norm(sample(testArray)-sample(testArray)) < R) # --> I already know there is an error here. What I'm trying to check is if coordinates of the sampled point agrees with restriction. i.e. I need to get the coordinates of points in RegularGrid. Could you please help me in this assert?
```
juliohm: Here you are confusing the index of the location with the coordinates of the location. The subsets returned by a partitioner are just indices of the locations in the domain. So when you `sample(testArray)` you are getting an integer and not a vector of coordinates. Please fix the test accordingly.

test/partitioning.jl
```diff
@@ -196,7 +196,31 @@
...
+
+    A = (1,1)
+    B = (10,10)
+    function SPred2(x,y)   # function to check if two points x and y are in a rectangle defined by points A and B
```
juliohm: This function is overly complicated. To check if a point `x` is inside the rectangle defined by corners `A` and `B`, you just need to do `all(A .<= x .<= B)`. So the entire function can be written as

```Julia
# check if points x and y belong to the rectangle [0.,10.]x[0.,10.]
pred(x, y) = all([0.,0.] .<= x .<=[10.,10.]) && all([0.,0.] .<= y .<= [10.,10.])
```

Notice that this may produce a result that is different than what you are expecting. What is the number of subsets you expect to see? Think before you run the code.

test/partitioning.jl
```diff
@@ -196,7 +196,31 @@
...
+      D = [0,0]
+      C[1] = min(A[1],B[1])
+      C[2] = min(A[2],B[2])
+      D[1] = max(A[1],B[1])
+      D[2] = max(A[2],B[2])
+      return (C[1]<=x[1]<=D[1]) & (C[2]<=x[2]<=D[2]) & (C[1]<=y[1]<=D[1]) & (C[2]<=y[2]<=D[2])
+    end
+
+    p = partition(g, SpatialPredicatePartitioner(SPred2))
+    s1 = subsets(p)
```
juliohm: Please follow the same convention as above:

```Julia
s = subsets(partition(g, p))
```

test/partitioning.jl
errearanhas: you meant to say `p = SpatialPredicatePartitioner((x, y) -> norm(x-y) < 5.0)`, and then `s = subsets(partition(g, p))`, right?

juliohm:
Yes

test/partitioning.jl
errearanhas: yeah! that's what I tried to say in
> I need to get the coordinates of points in RegularGrid

Is there a way to retrieve point coordinates from its integer index?

test/partitioning.jl
errearanhas: Thank you. Indeed way more elegant.

This example as it is here should generate just 1 partition. I've commited the wrong version of tuple B. The one I wanted to commit was B=(5,5).
This way, my test should check if there are 65 subsets, with only one of them having more than 1 point (in this case, 36 points exactly).

test/partitioning.jl
errearanhas: I belive that `getindex` method does what I need. Anyway, I changed strategy and commited another test inplace.

test/partitioning.jl
```diff
@@ -196,7 +196,20 @@
   end
 
   @testset "SpatialPredicatePartitioner" begin
-    # TODO
+    g = RegularGrid{Float64}(10,10)
+    
+    # check if each partition has only 1 point
```
juliohm: The test that follows does not check what the comment says. Please make sure that you check the size of each subset in the partition. I showed you before how to iterate over partitions:

```julia
for d in partition(g, sp)
  @test npoints(d) == 1
end
```

test/partitioning.jl
```diff
@@ -196,7 +196,20 @@
...
+    sp = SpatialPredicatePartitioner((x,y) -> norm(x-y) < 1.0)
+    s = subsets(partition(g, sp))
+    @test length(s) == 100
+
```
juliohm: Also, you had another test where you wanted to check if the points inside the subsets were indeed inside a ball of radius r. Can you please add it back?

test/partitioning.jl
```diff
@@ -196,7 +196,20 @@
...
+    # check if points x and y belong to the rectangle [0.,5.]x[0.,5.]
+    pred(x, y) = all([0.,0.] .<= x .<=[5.,5.]) && all([0.,0.] .<= y .<= [5.,5.])
+    sp = SpatialPredicatePartitioner(pred)
+    s = subsets(partition(g, sp))
+
+    @test length(s) == 65
```
juliohm: This test is not intuitive at first. Why 65? Can you please design a test with more intuition? Someone reading this later will have a hard time understanding the motivations of the test.

test/partitioning.jl
```diff
@@ -196,7 +196,20 @@
...
+    @test maximum(length.(s)) == 36
```
juliohm: Similar comment applies here. Please choose a better test or comment what your intuition is here so that others can follow.

test/partitioning.jl
juliohm: As discussed, the predicate must be transitive for that test to make sense, it is fine to leave it out.

test/partitioning.jl
juliohm: Yes, marking as solved.

test/partitioning.jl
juliohm: Marking as solved

PR #29, opened 2020-04-08
Excluded AppVeyor from visual tests.

evetion:


evetion:
Only some small errors left, because Int becomes and Int32 on x86 platforms and the tests assume Int64.

test/runtests.jl
```diff
@@ -14,9 +14,10 @@ ENV["GKSwstype"] = "100"
 # environment settings
 islinux = Sys.islinux()
 istravis = "TRAVIS" ∈ keys(ENV)
+isappveyor = "APPVEYOR" ∈ keys(ENV)
```
juliohm, line 17: I think we should create a third variable `CI = istravis || isappveyor`, and then use it in all places where I was using `istravis`.

test/runtests.jl
```diff
@@ -14,9 +14,10 @@ ENV["GKSwstype"] = "100"
...
 datadir = joinpath(@__DIR__,"data")
-visualtests = !istravis || (istravis && islinux)
-if !istravis
+visualtests = (!istravis || (istravis && islinux)) && !isappveyor
```
juliohm, line 19: This line would become `visualtests = !CI || (istravis && islinux)` meaning we do visual tests only locally or in Tavis+Linux where the plots are somewhat reproducible compared to local settings.

test/runtests.jl
```diff
@@ -14,9 +14,10 @@ ENV["GKSwstype"] = "100"
...
+if (!istravis && !isappveyor)
```
juliohm, line 20: This line would become `if !CI`

juliohm:
Strangely the tests are failing in Julia v1.1 due to MosaicViews.jl. I don't know if they've recently changed something that broke support in Julia v1.1, could you please check? I also wonder if we should actually support these versions of the language < 1.3. We've already dropped the LTS release 1.0, so 1.1 or 1.2 doesn't make much difference, agree?

evetion:
The previous tests were working. In the meantime `ImageCore` released a new version v0.8.13, which includes the latest MosaicViews.jl. We should report that somewhere upstream.

I agree that 1.1 and 1.2 matter less than the LTS release, but I mainly put in this work to actually test the supported versions + Windows. It's up to you what versions to support.

A final note, it seems that you yourself should go to ci.appveyor.com and enable this repository with your Github account (or JuliaEarth org). The ones we've used are in the JuliaGeo and my own account, but these are not automatically triggered, I started them manually.

evetion:
Made a fix for the newly introduced error: https://github.com/JuliaArrays/MosaicViews.jl/pull/16

juliohm:
Awesome 💯

PR #37, opened 2020-06-02
Update Project.toml

briochemc:
Just trying to bump compat with latest dataframes (not sure this'll work)


briochemc:
Well it seems this did not work! Nothing to see here 😅

PR #38, opened 2020-06-03
Fix axis range in recipe for RegularGrid

briochemc:


briochemc:
Trying to fix https://github.com/JuliaEarth/GeoStats.jl/issues/61

src/plotrecipes/domains/regular_grid.jl
```diff
@@ -23,6 +23,8 @@
   end
 end
 
+domainaxis(g::RegularGrid, dim) = range(g.origin[dim], step=g.spacing[dim], length=size(g)[dim])
```
juliohm, line 26: Nice. Could you use the public methods instead:

```julia
gridaxis(g::RegularGrid, dim) = range(origin(g)[dim], step=spacing(g)[dim], length=size(g)[dim])
```

Also given that this function is only called inside that `N==2` branch, perhaps we could just hard-code it there?

```julia
d = size(domain)
o = origin(domain)
s = spacing(domain)
x = range(o[1], step=s[1], length=d[1])
x = range(o[2], step=s[2], length=d[2])
```

What do you think?

briochemc: I don't know 🤷‍♂️  I really did a quick fix and tried to make it as minimal as possible, but I also thought you could use that function inside other pieces of code or other recipes as well? But please do as I think is best!

Re: testing, I wouldn't know how to test it really, except for a `@test plot(g) isa Plots.Plot`? (But that only checks that there is no error thrown, not that it does what we want it to do!)

juliohm: We use visual regression tests, you can add a line like this one here: https://github.com/JuliaEarth/GeoStatsBase.jl/blob/master/test/data.jl#L106-L110

Regarding the code snippet, what about we move the definitions

```julia
sz = size(domain)
or = origin(domain)
sp = spacing(domain)
```

to the top of the function right after the definition of `Z`, and then refer to these variables inside the branches like:

```julia
if N == 1
x = range(or[1], step=sp[1], length=sz[1])
x, Z
elseif N == 2
x = range(or[1], step=sp[1], length=sz[1])
y = range(or[2], step=sp[2], length=sz[2])
x, y, Z
elseif N == 3
...
end
```

that would be more self-contained :+1:

juliohm: I am calling the day here (BRT time), will the check the PR tomorrow. Thanks for the help.

briochemc: OK I will try but I make no promises! 😄

briochemc:
BTW, the current CI failure is not related to my tiny change, is it?

juliohm:
Probably not related to your change. More likely it is the usual Plots.jl noise. Whenever they release a new version the plots aesthetics change a bit and the visual regression tests fail. What I am doing is running these tests that fail every now and then locally to update the reference image stored in the test directory.

juliohm:
@briochemc could you please allow more commits on the PR? I think there is a box that you can mark to allow more commits. I have some changes that I would like to push before merging.

briochemc:
Sorry I missed this!

juliohm:
No worries. I found a time window during the day to work on it and preferred to move on quickly. Thanks again for catching this issue.

As a side note, I will likely do a lot of research with climate data in the next months. So issues like those with Haversine will be fixed more often. :+1:

PR #44, opened 2020-07-30
Kballsearcher

exepulveda:
Implementing KBallSearcher as we discussed. 
KBallSearcher uses knn and radius combined by searching for knn first and then filtering the resutls to locations withtin the radius.

### DIFF
--- a/src/neighborsearch/kball_searcher.jl
+++ b/src/neighborsearch/kball_searcher.jl
@@ -1,0 +1,56 @@
+# ------------------------------------------------------------------
+# Licensed under the ISC License. See LICENSE in the project root.
+#
+# Author: @exepulveda
+# ------------------------------------------------------------------
+
+"""
+    KBallSearcher(object, k, ball; locations)
+
+A method for searching neighbors in spatial `object` inside `neighborhood`.
+Used when doing estimation combining k-nearest and radius.
+It searches first for the k-nearest by using the internal kdtree and then
+limiting the results within the `BallNeighborhood`
+"""
+struct KBallSearcher{O,B,K} <: AbstractBoundedNeighborSearcher
+  # input fields
+  object::O
+  k::Int
+  ball::B
+  locations::Vector{Int}
+
+  # state fields
+  kdtree::K
+end
+
+
+function KBallSearcher(object::O, k::Int, ball::B; locations=1:npoints(object)) where {O,B}
+  @assert 1 ≤ k ≤ length(locations) "number of neighbors must be smaller than number of locations"
+  @assert locations ⊆ 1:npoints(object) "invalid locations for spatial object"
+  kdtree = KDTree(coordinates(object, locations), metric(ball))
+  KBallSearcher{O,B,typeof(kdtree)}(object, k, ball, locations, kdtree)
+end
+
+maxneighbors(searcher::KBallSearcher) = searcher.k
+
+function search!(neighbors::AbstractVector{Int}, xₒ::AbstractVector,
+                 searcher::KBallSearcher; mask=nothing)
+  k    = searcher.k
+  ball = searcher.ball
+  r    = radius(ball)
+
+  inds, dists = knn(searcher.kdtree, xₒ, k, true)
+  locs = view(searcher.locations, inds)
+
+  nneigh = 0
+  @inbounds for i in 1:length(inds)
+    #if mask is true, the locations must be skipped
+    useit = (mask === nothing) || !mask[locs[i]]
+    if useit && (dists[i] ≤ r)
+      nneigh += 1
+      neighbors[nneigh] = locs[i]
+    end
+  end
+
+  nneigh
+end
--- a/src/neighborsearch/kball_searcher.jl
+++ b/src/neighborsearch/kball_searcher.jl
@@ -1,16 +1,12 @@
 # ------------------------------------------------------------------
 # Licensed under the ISC License. See LICENSE in the project root.
-#
-# Author: @exepulveda
 # ------------------------------------------------------------------
 
 """
     KBallSearcher(object, k, ball; locations)
 
-A method for searching neighbors in spatial `object` inside `neighborhood`.
-Used when doing estimation combining k-nearest and radius.
-It searches first for the k-nearest by using the internal kdtree and then
-limiting the results within the `BallNeighborhood`
+A method that searches `k` nearest neighbors and then filters
+these neighbors using a norm `ball`.
 """
 struct KBallSearcher{O,B,K} <: AbstractBoundedNeighborSearcher
   # input fields
@@ -35,18 +31,21 @@ maxneighbors(searcher::KBallSearcher) = searcher.k
 
 function search!(neighbors::AbstractVector{Int}, xₒ::AbstractVector,
                  searcher::KBallSearcher; mask=nothing)
-  k    = searcher.k
-  ball = searcher.ball
-  r    = radius(ball)
+  k = searcher.k
+  r = radius(searcher.ball)
 
   inds, dists = knn(searcher.kdtree, xₒ, k, true)
   locs = view(searcher.locations, inds)
 
+  # keep neighbors inside ball
+  keep = dists .≤ r
+
+  # possibly mask some of the neighbors
+  isnothing(mask) || (keep .*= mask[locs])
+
   nneigh = 0
-  @inbounds for i in 1:length(inds)
-    #if mask is true, the locations must be skipped
-    useit = (mask === nothing) || !mask[locs[i]]
-    if useit && (dists[i] ≤ r)
+  @inbounds for i in 1:k
+    if keep[i]
       nneigh += 1
       neighbors[nneigh] = locs[i]
     end
--- a/test/neighborsearch.jl
+++ b/test/neighborsearch.jl
@@ -1,0 +1,53 @@
+@testset "Neighborsearch" begin
+  @testset "KBallSearcher" begin
+    #create a regular grid to test
+    domain = RegularGrid(10,10)
+
+    #test 1: all points should be found
+    k = 10
+    ball = BallNeighborhood(100.0)
+    kballsearcher = KBallSearcher(domain, k, ball)
+
+    xₒ = [5.0,5.0]
+    neighbors = search(xₒ, kballsearcher)
+
+    @test length(neighbors) == 10
+
+    #test 2: radius of 1.0 m 5 points should be found
+    k = 10
+    ball = BallNeighborhood(1.0)
+    kballsearcher = KBallSearcher(domain, k, ball)
+
+    xₒ = [5.0,5.0]
+    neighbors = search(xₒ, kballsearcher)
+    @test length(neighbors) == 5
+    @test neighbors[1] == 56
+
+    #testing with masks
+    mask = falses(npoints(domain))
+    mask[56] = true #excluding location #56
+    neighbors = search(xₒ, kballsearcher;mask=mask)
+    @test length(neighbors) == 4
+
+    #test 3: radius of 1.0 m but far just to find 1 point
+    k = 10
+    ball = BallNeighborhood(1.0)
+    kballsearcher = KBallSearcher(domain, k, ball)
+
+    xₒ = [-0.2,-0.2]
+    neighbors = search(xₒ, kballsearcher)
+
+    @test length(neighbors) == 1
+
+    #test 4: radius of 1.0 m but far just to find no points
+    k = 10
+    ball = BallNeighborhood(1.0)
+    kballsearcher = KBallSearcher(domain, k, ball)
+
+    xₒ = [-10.0,-10.0]
+    neighbors = search(xₒ, kballsearcher)
+
+    @test length(neighbors) == 0
+    
+  end
+end
--- a/test/neighborsearch.jl
+++ b/test/neighborsearch.jl
@@ -6,48 +6,47 @@
     #test 1: all points should be found
     k = 10
     ball = BallNeighborhood(100.0)
-    kballsearcher = KBallSearcher(domain, k, ball)
+    searcher = KBallSearcher(domain, k, ball)
 
     xₒ = [5.0,5.0]
-    neighbors = search(xₒ, kballsearcher)
+    neighbors = GeoStatsBase.search(xₒ, searcher)
 
     @test length(neighbors) == 10
 
     #test 2: radius of 1.0 m 5 points should be found
     k = 10
     ball = BallNeighborhood(1.0)
-    kballsearcher = KBallSearcher(domain, k, ball)
+    searcher = KBallSearcher(domain, k, ball)
 
     xₒ = [5.0,5.0]
-    neighbors = search(xₒ, kballsearcher)
+    neighbors = GeoStatsBase.search(xₒ, searcher)
     @test length(neighbors) == 5
     @test neighbors[1] == 56
 
     #testing with masks
-    mask = falses(npoints(domain))
-    mask[56] = true #excluding location #56
-    neighbors = search(xₒ, kballsearcher;mask=mask)
+    mask = trues(npoints(domain))
+    mask[56] = false #excluding location #56
+    neighbors = GeoStatsBase.search(xₒ, searcher, mask=mask)
     @test length(neighbors) == 4
 
     #test 3: radius of 1.0 m but far just to find 1 point
     k = 10
     ball = BallNeighborhood(1.0)
-    kballsearcher = KBallSearcher(domain, k, ball)
+    searcher = KBallSearcher(domain, k, ball)
 
     xₒ = [-0.2,-0.2]
-    neighbors = search(xₒ, kballsearcher)
+    neighbors = GeoStatsBase.search(xₒ, searcher)
 
     @test length(neighbors) == 1
 
     #test 4: radius of 1.0 m but far just to find no points
     k = 10
     ball = BallNeighborhood(1.0)
-    kballsearcher = KBallSearcher(domain, k, ball)
+    searcher = KBallSearcher(domain, k, ball)
 
     xₒ = [-10.0,-10.0]
-    neighbors = search(xₒ, kballsearcher)
+    neighbors = GeoStatsBase.search(xₒ, searcher)
 
     @test length(neighbors) == 0
-    
   end
 end
--- a/src/GeoStatsBase.jl
+++ b/src/GeoStatsBase.jl
@@ -226,6 +226,7 @@ export
   NearestNeighborSearcher,
   NeighborhoodSearcher,
   BoundedSearcher,
+  KBallSearcher,
   search!, search,
   maxneighbors,
   object,
--- a/src/domains/point_set.jl
+++ b/src/domains/point_set.jl
@@ -9,6 +9,20 @@ A set of points with coordinate matrix `coords`. The number of rows
 of the matrix is the dimensionality of the domain whereas the number
 of columns is the number of points in the set. Alternatively, `coords`
 can be a vector of tuples (i.e. points).
+
+## Examples
+
+Create a 2D point set with 100 points:
+
+```julia
+julia> PointSet(rand(2,100))
+```
+
+Or equivalently, using a vector of tuples:
+
+```julia
+julia> PointSet([(rand(),rand()) for i in 1:100])
+```
 """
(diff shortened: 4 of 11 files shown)


src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,52 @@
+# ------------------------------------------------------------------
+# Licensed under the ISC License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+KBallSearcher(object, neighborhood)
```
juliohm, line 6: Could you please prepend 4 whitespaces to this line? This way the Julia help prompt will highlight the function signature.

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,52 @@
...
+  searcher::S
+  nmax::Int
+end
+
+object(searcher::KBallSearcher) = object(searcher.searcher)
+
+maxneighbors(searcher::KBallSearcher) = searcher.nmax
+
+function search!(neighbors::AbstractVector{Int}, xₒ::AbstractVector,
+  searcher::KBallSearcher; mask=nothing)
```
juliohm, line 23: [Pedantic mode activated]
Could you please align this line to fit below the first argument (in this case the "n" of "neighbors")
[Pedantic mode deactivated]

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,52 @@
...
+"""
+KBallSearcher(object, neighborhood)
+
+A method for searching neighbors in spatial `object` inside `neighborhood`.
+Used when doing estimation combining k-nearest and radius.
+It searches first for the k-nearest by using the internal kdtree and then
+limiting the results within the `BallNeighborhood` by calling `isneighbor`
+"""
+struct KBallSearcher{S<:AbstractNeighborSearcher} <: AbstractBoundedNeighborSearcher
+  searcher::S
```
juliohm, line 14: Should we make use of a more restrictive `S <: NearestNeighborSearcher`? I think the idea of this searcher only works with nearest-neighbor search as the inner searcher right?

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,52 @@
...
+function search!(neighbors::AbstractVector{Int}, xₒ::AbstractVector,
+  searcher::KBallSearcher; mask=nothing)
+
+  knn_searcher = searcher.searcher #internar nearest searcher
+  k = searcher.nmax
+  neigh = knn_searcher.neigh       #the Ball Neighborhood
+  object = knn_searcher.object
+  radius = neigh.radius
+
+  inds, dists = knn(knn_searcher.kdtree, xₒ, k, true)
```
juliohm, line 31: I think what you really want to do here is leverage the inner searcher and call the `search!` method for it?

```julia
locs = search(xₒ, searcher.searcher, mask=mask)
```
And then post-process the locations with `isneighbor`?

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,52 @@
...
+Used when doing estimation combining k-nearest and radius.
+It searches first for the k-nearest by using the internal kdtree and then
+limiting the results within the `BallNeighborhood` by calling `isneighbor`
+"""
+struct KBallSearcher{S<:AbstractNeighborSearcher} <: AbstractBoundedNeighborSearcher
+  searcher::S
+  nmax::Int
+end
+
+object(searcher::KBallSearcher) = object(searcher.searcher)
```
juliohm, line 18: I think this isn't needed if you just call the `search` method of the inner `searcher` directly. See comment bellow.

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,52 @@
...
+KBallSearcher(object, neighborhood)
+
+A method for searching neighbors in spatial `object` inside `neighborhood`.
+Used when doing estimation combining k-nearest and radius.
+It searches first for the k-nearest by using the internal kdtree and then
+limiting the results within the `BallNeighborhood` by calling `isneighbor`
+"""
+struct KBallSearcher{S<:AbstractNeighborSearcher} <: AbstractBoundedNeighborSearcher
+  searcher::S
+  nmax::Int
```
juliohm, line 15: Maybe the second field of the KBallSearcher is the ball? So you first search with the inner searcher, and then post-process the results with the ball, makes sense? No need to specify `nmax` because that is already inside the inner nearest-neighbor searcher.

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,52 @@
...
+    if masked 
+      coordinates!(x, object, loc)
+      if isneighbor(neigh, xₒ, x) #limiting to the radius
+        #this may be improve as the knn already returns the distances, but less generic 
+        #because the BallNeighborhood can use a different metric indeed
+        nneigh += 1
+        neighbors[nneigh] = loc
+      end
+    end
+  end
```
juliohm, line 49: This post-processing can be simplified after the other changes above take place.

test/neighborsearch.jl
```diff
@@ -0,0 +1,57 @@
...
+    # KBallSearcher
+
+    #create a regular grid to test
+    pdomain = RegularGrid(10,10)
+
+    #test 1: all points should be found
+    maxneighbors = 10
+    neigh = BallNeighborhood(100.0)
+    searcher  = NeighborhoodSearcher(pdomain, neigh)
+    kballsearcher = KBallSearcher(searcher,maxneighbors)
```
juliohm, line 12: So the example I had in mind when you explained the KBallSearcher concept was something like:

```julia
knnsearcher = NearestNeighborSearcher(pdomain, maxneighbors)
kballsearcher = KBallSearcher(knnsearcher, neigh)
```
So the inner searcher is always a `NearestNeighborSearcher` and the other missing step is the `neigh` (ball). Makes sense?

test/neighborsearch.jl
```diff
@@ -0,0 +1,57 @@
...
+    xₒ = [5.0,5.0]
+    nneigh = search!(neighbors,xₒ,kballsearcher; mask=nothing)
+
+    @test nneigh == 5
+
+    #test 3: radius of 1.0 m but far just to find 1 point
+    maxneighbors = 10
+    neigh = BallNeighborhood(1.0)
+    searcher  = NeighborhoodSearcher(pdomain, neigh)
+    kballsearcher = KBallSearcher(searcher,maxneighbors)
```
juliohm, line 36: Previous comment applies here too.

test/neighborsearch.jl
```diff
@@ -0,0 +1,57 @@
...
+    xₒ = [-0.2,-0.2]
+    nneigh = search!(neighbors,xₒ,kballsearcher; mask=nothing)
+
+    @test nneigh == 1
+
+    #test 4: radius of 1.0 m but far just to find no points
+    maxneighbors = 10
+    neigh = BallNeighborhood(1.0)
+    searcher  = NeighborhoodSearcher(pdomain, neigh)
+    kballsearcher = KBallSearcher(searcher,maxneighbors)
```
juliohm, line 48: And here as well.

juliohm:
Just in case you are not 100% familiar with the GitHub workflow, if you push new commits to the same branch on your fork, the PR will be automatically updated.

src/neighborsearch/kball_searcher.jl
exepulveda: We could do that, but in that way we would need to recalculate distances (search function does not return distances that has already done by the inner kdtree). That is the reason for implementing this as it is.

src/neighborsearch/kball_searcher.jl
exepulveda: That is one option. I just tried to reuse as much as possible what is done. For user perspective, it could be easier to just indicate something like this:
solver = Kriging(
    :cu => (variogram=vmodel,maxneighbors=25,radius=50.0)
)
but I guess we don't want to change the API.

src/neighborsearch/kball_searcher.jl
juliohm: What we can do then is consider `KBallSearcher` its own implementation without relying on other existing searchers. In this case, it could have two fields `ball` and `k`. The constructor instantiates the kdtree similar to what is done in the NearestNeighborSearcher constructor, and then we implement the search explicitly without an inner searcher. What do you think?

src/neighborsearch/kball_searcher.jl
juliohm: Yes, I think it is more general to leave the option as neighborhood because then we have other kinds of neighborhood that could be useful that are not norm-balls.

src/neighborsearch/kball_searcher.jl
exepulveda: Much better I think. For example:
` Kriging(
:cu => (variogram=vmodel,searcher=KBallSearcher(k=25,radius=50.0))
)`

juliohm: I think the Kriging interface is another issue that we could consider later. Let's first work on this PR to get the KBallSearcher in place, then we proceed to the other PR to figure out an optimal set of parameters 👍

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,68 @@
...
+
+
+function KBallSearcher(object::O, k::Int, ball::BallNeighborhood{T,M}; locations=1:npoints(object)) where {O,T,M}
+  @assert 1 ≤ k ≤ length(locations) "number of neighbors must be smaller than number of locations"
+  @assert locations ⊆ 1:npoints(object) "invalid locations for spatial object"
+  kdtree = KDTree(coordinates(object, locations), ball.metric)
+  KBallSearcher{O,typeof(kdtree),T,M}(object, k, ball, locations, kdtree)
+end
+
+maxneighbors(searcher::KBallSearcher) = searcher.nmax
```
juliohm, line 36: I think this is outdated code, right?

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,68 @@
...
+  # state fields
+  kdtree::K
+
+end
+
+
+function KBallSearcher(object::O, k::Int, ball::BallNeighborhood{T,M}; locations=1:npoints(object)) where {O,T,M}
+  @assert 1 ≤ k ≤ length(locations) "number of neighbors must be smaller than number of locations"
+  @assert locations ⊆ 1:npoints(object) "invalid locations for spatial object"
+  kdtree = KDTree(coordinates(object, locations), ball.metric)
```
juliohm, line 32: We try to avoid accessing fields of objects directly. In this case we have `metric(ball)` defined. Could you please use that instead?

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,68 @@
...
+
+"""
+    KBallSearcher(object, k, ball; locations)
+
+A method for searching neighbors in spatial `object` inside `neighborhood`.
+Used when doing estimation combining k-nearest and radius.
+It searches first for the k-nearest by using the internal kdtree and then
+limiting the results within the `BallNeighborhood`
+"""
+struct KBallSearcher{O,K,T,M} <: AbstractBoundedNeighborSearcher
```
juliohm, line 15: I think we can simplify the type parameters here. I've learned over time that we sometimes parameterize things too heavily given our C++ background. The idea is to parameterize with the most generic types possible so that the code works with types that we didn't imagine at first:

```julia
struct KBallSearcher{O,B,K} <: AbstractBoundedNeighborSearcher
  object::O
  k::Int
  ball::B
  locations::Vector{Int}
  
  # state fields
  kdtree::K
end
```

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,68 @@
...
+  ball::BallNeighborhood{T,M}
+  locations::Vector{Int}
+
+  # state fields
+  kdtree::K
+
+end
+
+
+function KBallSearcher(object::O, k::Int, ball::BallNeighborhood{T,M}; locations=1:npoints(object)) where {O,T,M}
```
juliohm, line 29: We can simplify this signature as well to something like:

```julia
function KBallSearcher(object::O, k::Int, ball::B; locations=1:npoints(object)) where {O,B}
  # ...
  KBallSearcher{O,B,typeof(kdtree)}(object, k, ball, locations, kdtree)
end
```

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,68 @@
...
+end
+
+maxneighbors(searcher::KBallSearcher) = searcher.nmax
+
+function search!(neighbors::AbstractVector{Int}, xₒ::AbstractVector,
+                 searcher::KBallSearcher; mask=nothing)
+
+  k      = searcher.k
+  ball   = searcher.ball
+  radius = ball.radius
```
juliohm, line 43: Please use `radius(ball)` instead

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,68 @@
...
+
+  inds, dists = knn(searcher.kdtree, xₒ, k, true)
+  locs = view(searcher.locations, inds) #translate from relative indices (inds) to pdomain indices
```
juliohm, line 46: Do you mind removing the comment? I think it helps somehow but it also can be inferred from the constructor easily. Also it contains this `pdomain` jargon people may not follow :relaxed:

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,68 @@
...
+
+  if mask ≠ nothing
+    nneigh = 0
+    @inbounds for i in 1:length(inds)
```
juliohm, line 50: Maybe replace `length(inds)` by `k`

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,68 @@
...
+      #if mask is true, the locations is already
```
juliohm, line 51: mind removing this comment as well?

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,68 @@
...
+      if mask[locs[i]] && (dists[i] ≤ radius)
+        nneigh += 1
+        neighbors[nneigh] = locs[i]
+      end
+    end
+  else
+    nneigh = 0
+    @inbounds for i in 1:length(inds)
```
juliohm, line 59: replace `length(inds)` by `k`

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,68 @@
...
+      if dists[i] ≤ radius
+        nneigh += 1
+        neighbors[nneigh] = locs[i]
+      end
+    end
+  end
+
+  nneigh
+end
```
juliohm, line 68: I never seen GitHub add this newline warning in a review, interesting. Do you think we should add a new line like in other files?

src/neighborsearch/kball_searcher.jl
```diff
@@ -0,0 +1,68 @@
...
+                 searcher::KBallSearcher; mask=nothing)
+
+  k      = searcher.k
+  ball   = searcher.ball
+  radius = ball.radius
+
+  inds, dists = knn(searcher.kdtree, xₒ, k, true)
+  locs = view(searcher.locations, inds) #translate from relative indices (inds) to pdomain indices
+
+  if mask ≠ nothing
```
juliohm, line 48: This `mask` business I've introduced some time ago needs to be refactored at some point in the future. The implementation feels clumsy because of this mask, yet I don't have any idea at the moment to replace this feature by something clever. Let me know if you have something in mind.

test/neighborsearch.jl
```diff
@@ -0,0 +1,59 @@
+@testset "Neighborsearch" begin
+  @testset "KBallSearcher" begin
+    # KBallSearcher
+
```
juliohm, line 4: Can erase this comment as we have the testset already :+1:

test/neighborsearch.jl
```diff
@@ -0,0 +1,59 @@
...
+    #create a regular grid to test
+    pdomain = RegularGrid(10,10)
```
juliohm, line 6: I would call this `D` or `dom` or something like it. The `pdomain` jargon was introduced in some solvers just to mean "problem domain". Here it doesn't feel very natural.

test/neighborsearch.jl
```diff
@@ -0,0 +1,59 @@
...
+
+    #test 1: all points should be found
+    k = 10
+    ball = BallNeighborhood(100.0)
+    kballsearcher = KBallSearcher(pdomain, k, ball)
+
+    neighbors = neighbors = Vector{Int}(undef, k)
```
juliohm, line 13: Could you please double check this neighbors = neighbors here?

test/neighborsearch.jl
```diff
@@ -0,0 +1,59 @@
...
+    xₒ = [5.0,5.0]
+    nneigh = search!(neighbors,xₒ,kballsearcher; mask=nothing)
```
juliohm, line 15: you can omit the mask argument since it defaults to nothing, it will make code cleaner and if we find a better strategy than this mask in the future the test will still work.

juliohm: also instead of testing the low-level `search!` here you could be testing the allocating version `search` that has a default implementation calling `search!`. You can just do `neighbors = search(xo, kballsearcher)`

test/neighborsearch.jl
```diff
@@ -0,0 +1,59 @@
...
+    @test nneigh == 10
+
+    #test 2: radius of 1.0 m 5 points should be found
+    k = 10
+    ball = BallNeighborhood(1.0)
+    kballsearcher = KBallSearcher(pdomain, k, ball)
+
+    neighbors = Vector{Int}(undef, k)
+    xₒ = [5.0,5.0]
+    nneigh = search!(neighbors,xₒ,kballsearcher; mask=nothing)
```
juliohm, line 26: sample applies to here, you could do `neighbors = search(xo, kballsearcher)` and then `nneigh = length(neighbors)`

src/neighborsearch/kball_searcher.jl
exepulveda: Sure. Could we explain me the reasons?

juliohm: It is because the object `ball` has a publicly defined API that we try to use. At some point in the future if the field name changes to something else, we will still be able to call `metric(ball)`. We only access fields directly when we are implementing a method for a function that is related to the struct we are defining the behavior. For example, in this case we are implementing the `searcher` so we can access the fields directly. However, the field `ball` of the searcher has its behavior defined elsewhere in the source and so we should only rely on the public api. Makes sense?

juliohm:
@exepulveda I remembered this issue today, and I think we may be overcomplicating the solution? We already have two searchers: (1) k-nearest neighbors and (2) ball neighborhood. The neighborhood searcher returns all the locations inside the ball, no matter the number. The problem in the Kriging solver is that we are fitting this neighborhood searcher to the entire `pdomain` as opposed to the `pdata` locations. If we construct the searcher with `pdata` we solve the issue almost completely. The only missing part is the treatment of `minneighbors` and `maxneighbors` parameters. We could follow your suggestion and leave the cases with `n < minneighbors` untouched, and randomly select `maxneighbors` from the pool of `n` without replacement to make sure that we never exceed this value. What do you think?

In summary, the fixes that we could implement quickly in Kriging are:

- Fit the existing neighborhood searcher to `pdata`
- In the Kriging loop, get all the `n` neighbors
- If `n < minneighbors`, then place `missing` at location
- Otherwise sample `min(n, maxneighbors)` locations without replacement

If we do this, we have a faster Kriging already with neighborhoods that we could iterate more later with more improvements. This would also avoid introducing a new searcher like in this PR.

exepulveda:
That is partially correct. Remember that (1) and (2) can be combined and it will work faster (that was the original idea with the new searcher). I agree with indexing pdata instead will fix the issue in 95% roughly. If you don't mind let me try to do it. I can at the same time learn the inside of GeoStatsBase.

juliohm:
I added a few more fixes to the implementation, and by accident also added other commits from the master branch. Sorry for polluting the PR, I will merge and reconcile in the master branch. :+1:

PR #48, opened 2020-08-15
Add PredicateFilter

juliohm:
Fixes https://github.com/JuliaEarth/GeoStats.jl/issues/101

We provide a new `PredicateFilter` that acts on the "rows" of the spatial data table. The PR also introduces a convenience function `filter(pred, sdata)` to follow Julia's builtin syntax.


juliohm:
I will merge this one in order to fix other issues downstream :+1:

PR #49, opened 2020-08-19
Add inside operation

juliohm:
This PR introduces a new geometric operation called `inside` that takes a spatial object and returns a subset inside a given geometry. After we have this function properly tested, we can introduce a convenient function `slice` that calls `inside` with a `Rectangle`.

Fix https://github.com/JuliaEarth/GeoStats.jl/issues/72


juliohm:
This is ready for review @exepulveda. After this PR is merged, we can start a new one to introduce a `slice` function :+1:

PR #52, opened 2020-09-01
Refact pointset

exepulveda:
With this change some improvements are observed.

Current implementation using Matrix:
```

for i in 1:10
  solution = @time solve(problem, solver)
end
  4.042634 seconds (3.67 M allocations: 11.620 GiB, 6.44% gc time)
  4.037391 seconds (3.67 M allocations: 11.620 GiB, 5.93% gc time)
  4.064429 seconds (3.67 M allocations: 11.620 GiB, 5.65% gc time)
  4.116754 seconds (3.67 M allocations: 11.620 GiB, 5.33% gc time)
  4.125295 seconds (3.67 M allocations: 11.620 GiB, 5.17% gc time)
  4.235886 seconds (3.67 M allocations: 11.620 GiB, 8.01% gc time)
  4.095777 seconds (3.67 M allocations: 11.620 GiB, 4.67% gc time)
  4.114237 seconds (3.67 M allocations: 11.620 GiB, 4.70% gc time)
  4.185361 seconds (3.67 M allocations: 11.620 GiB, 4.85% gc time)
  4.126609 seconds (3.67 M allocations: 11.620 GiB, 4.42% gc time)
```

With the proposed implementation:

```
for i in 1:10
         @time solve(problem, solver)
       end
  3.248977 seconds (3.21 M allocations: 3.808 GiB, 3.70% gc time)
  3.346375 seconds (3.21 M allocations: 3.808 GiB, 3.55% gc time)
  3.371415 seconds (3.21 M allocations: 3.808 GiB, 3.62% gc time)
  3.399316 seconds (3.21 M allocations: 3.808 GiB, 3.56% gc time)
  3.378670 seconds (3.21 M allocations: 3.808 GiB, 3.47% gc time)
  3.336218 seconds (3.21 M allocations: 3.808 GiB, 3.49% gc time)
  3.237753 seconds (3.21 M allocations: 3.808 GiB, 3.42% gc time)
  3.343776 seconds (3.21 M allocations: 3.808 GiB, 5.58% gc time)
  3.266192 seconds (3.21 M allocations: 3.808 GiB, 3.17% gc time)
  3.287763 seconds (3.21 M allocations: 3.808 GiB, 3.37% gc time)
```

All tests (locally) passed OK.

### DIFF
--- a/src/domains/point_set.jl
+++ b/src/domains/point_set.jl
@@ -3,15 +3,11 @@
 # ------------------------------------------------------------------
 
 """
-    PointSet(coords)
+PointSet(coords)
 
-A set of points with coordinates vector `coords`. The vector
-contains many coordinates (number of points), where each coordinate
-is implemented by a static array of length same as the dimensionality.
-To build the PointSet a Matrix can be use. The matrix is the dimensionality 
-of the domain whereas the number of columns is the number of points in the set
-or a Vector of tuples that will be converted to a Vector of static array.
-
+A set of points with coordinates coords. Each point is represented by a static vector
+or tuple. Alternatively, coords can be a matrix where the number of rows equals the
+number of dimensions.
 ## Examples
 
 Create a 2D point set with 100 points:
@@ -28,47 +24,36 @@ julia> PointSet([(rand(),rand()) for i in 1:100])
 """
 struct PointSet{T,N} <: SpatialDomain{T,N}
   coords::Vector{SVector{N,T}} 
-
-  function PointSet{T,N}(coords) where {N,T}
-    @assert !isempty(coords) "coordinates must be non-empty"
-    new(coords)
-  end
 end
 
-function PointSet(coords::AbstractMatrix{T}) where {T}
-  N = size(coords, 1)
-  points = Vector{SVector{N,T}}()
-  for row in eachcol(coords)
-    push!(points,SVector{N,T}(row))
-  end
-  PointSet{T,N}(points)
+PointSet(coords::AbstractMatrix) = PointSet(collect(eachcol(coords)))
+PointSet(coords::AbstractVector{<:NTuple}) = PointSet(SVector.(coords))
+
+function PointSet(coords::AbstractVector{<:AbstractVector})
+  @assert !isempty(coords) "coordinates must be non-empty"
+  @assert length(coords) > 0 "coordinates must contain one point at least"
+  N = length(coords[1]) #getting the dimensionality of the point
+  PointSet(SVector{N}.(coords))
 end
 
-function PointSet(coords::AbstractVector{NTuple{N,T}}) where {N,T}
-  points = Vector{SVector{N,T}}()
-  for row in coords
-    push!(points,SVector{N,T}(row))
-  end
-  PointSet{T,N}(points)
-end
-
-nelms(ps::PointSet) = size(ps.coords,1)
+nelms(ps::PointSet) = length(ps.coords)
 
 function coordinates!(buff::AbstractVector{T}, ps::PointSet{T,N},
-                      location::Int) where {N,T}
+                      ind::Int) where {N,T}
   @inbounds for i in 1:N
-    buff[i] = ps.coords[location][i]
+    buff[i] = ps.coords[ind][i]
   end
 end
 # ------------
 # IO methods
 # ------------
 function Base.show(io::IO, ps::PointSet{T,N}) where {N,T}
-  npts = nelms(ps)
+  npts = length(ps.coords)
   print(io, "$npts PointSet{$T,$N}")
 end
 
 function Base.show(io::IO, ::MIME"text/plain", ps::PointSet{T,N}) where {N,T}
   println(io, ps)
-  Base.print_array(io, ps.coords)
+  m = hcat(ps.coords...)
+  Base.print_array(io, m)
 end
--- a/src/domains/point_set.jl
+++ b/src/domains/point_set.jl
@@ -5,10 +5,12 @@
 """
     PointSet(coords)
 
-A set of points with coordinate matrix `coords`. The number of rows
-of the matrix is the dimensionality of the domain whereas the number
-of columns is the number of points in the set. Alternatively, `coords`
-can be a vector of tuples (i.e. points).
+A set of points with coordinates vector `coords`. The vector
+contains many coordinates (number of points), where each coordinate
+is implemented by a static array of length same as the dimensionality.
+To build the PointSet a Matrix can be use. The matrix is the dimensionality 
+of the domain whereas the number of columns is the number of points in the set
+or a Vector of tuples that will be converted to a Vector of static array.
 
 ## Examples
 
@@ -25,7 +27,8 @@ julia> PointSet([(rand(),rand()) for i in 1:100])
 ```
 """
 struct PointSet{T,N} <: SpatialDomain{T,N}
-  coords::Matrix{T}
+  # coords::Matrix{T}
+  coords::Vector{SVector{N,T}} 
 
   function PointSet{T,N}(coords) where {N,T}
     @assert !isempty(coords) "coordinates must be non-empty"
@@ -33,26 +36,36 @@ struct PointSet{T,N} <: SpatialDomain{T,N}
   end
 end
 
-PointSet(coords::AbstractMatrix{T}) where {T} =
-  PointSet{T,size(coords,1)}(coords)
+function PointSet(coords::AbstractMatrix{T}) where {T}
+  N = size(coords, 1)
+  points = Vector{SVector{N,T}}()
+  for row in eachcol(coords)
+    push!(points,SVector{N,T}(row))
+  end
+  PointSet{T,N}(points)
+end
 
-PointSet(coords::AbstractVector{NTuple{N,T}}) where {N,T} =
-  PointSet([c[i] for i in 1:N, c in coords])
+function PointSet(coords::AbstractVector{NTuple{N,T}}) where {N,T}
+  points = Vector{SVector{N,T}}()
+  for row in coords
+    push!(points,SVector{N,T}(row))
+  end
+  PointSet{T,N}(points)
+end
 
-nelms(ps::PointSet) = size(ps.coords, 2)
+nelms(ps::PointSet) = size(ps.coords,1)
 
 function coordinates!(buff::AbstractVector{T}, ps::PointSet{T,N},
                       location::Int) where {N,T}
   @inbounds for i in 1:N
-    buff[i] = ps.coords[i,location]
+    buff[i] = ps.coords[location][i]
   end
 end
-
 # ------------
 # IO methods
 # ------------
 function Base.show(io::IO, ps::PointSet{T,N}) where {N,T}
-  npts = size(ps.coords, 2)
+  npts = nelms(ps)
   print(io, "$npts PointSet{$T,$N}")
 end
 
--- a/src/domains/point_set.jl
+++ b/src/domains/point_set.jl
@@ -27,7 +27,6 @@ julia> PointSet([(rand(),rand()) for i in 1:100])
 ```
 """
 struct PointSet{T,N} <: SpatialDomain{T,N}
-  # coords::Matrix{T}
   coords::Vector{SVector{N,T}} 
 
   function PointSet{T,N}(coords) where {N,T}
--- a/src/domains/point_set.jl
+++ b/src/domains/point_set.jl
@@ -26,15 +26,8 @@ struct PointSet{T,N} <: SpatialDomain{T,N}
   coords::Vector{SVector{N,T}} 
 end
 
-PointSet(coords::AbstractMatrix) = PointSet(collect(eachcol(coords)))
+PointSet(coords::AbstractMatrix) = PointSet(collect(SVector{size(coords,1)}.(eachcol(coords))))
 PointSet(coords::AbstractVector{<:NTuple}) = PointSet(SVector.(coords))
-
-function PointSet(coords::AbstractVector{<:AbstractVector})
-  @assert !isempty(coords) "coordinates must be non-empty"
-  @assert length(coords) > 0 "coordinates must contain one point at least"
-  N = length(coords[1]) #getting the dimensionality of the point
-  PointSet(SVector{N}.(coords))
-end
 
 nelms(ps::PointSet) = length(ps.coords)
 
--- a/src/domains/point_set.jl
+++ b/src/domains/point_set.jl
@@ -3,11 +3,12 @@
 # ------------------------------------------------------------------
 
 """
-PointSet(coords)
+    PointSet(coords)
 
-A set of points with coordinates coords. Each point is represented by a static vector
-or tuple. Alternatively, coords can be a matrix where the number of rows equals the
+A set of points with coordinates `coords`. Each point is represented by a static vector
+or tuple. Alternatively, `coords` can be a matrix where the number of rows equals the
 number of dimensions.
+
 ## Examples
 
 Create a 2D point set with 100 points:
@@ -47,6 +48,6 @@ end
 
 function Base.show(io::IO, ::MIME"text/plain", ps::PointSet{T,N}) where {N,T}
   println(io, ps)
-  m = hcat(ps.coords...)
+  m = reduce(hcat, ps.coords) #hcat(ps.coords...)
   Base.print_array(io, m)
 end
--- a/test/domains.jl
+++ b/test/domains.jl
@@ -6,7 +6,7 @@
     @test coordinates(ps, 2) == [0., 1.]
 
     @test sprint(show, ps) == "2 PointSet{Float64,2}"
-    @test sprint(show, MIME"text/plain"(), ps) == "2 PointSet{Float64,2}\n 1.0  0.0\n 0.0  1.0"
+    @test sprint(show, MIME"text/plain"(), ps) == "2 PointSet{Float64,2}\n [1.0, 0.0]\n [0.0, 1.0]"
 
     if visualtests
       Random.seed!(2019)
--- a/test/domains.jl
+++ b/test/domains.jl
@@ -1,25 +1,33 @@
 @testset "Domains" begin
   @testset "PointSet" begin
-    # same values, different representations
+    # testing different ways to build a PointSet results in the same PointSet
     ps1 = PointSet([1. 0.; 0. 1.])            # as matrix
     ps2 = PointSet([(1.0, 0.0), (0.0, 1.0)])  # as vector of tuples
-    for ps in [ps1, ps2]
-      @test nelms(ps) == 2
-      @test coordinates(ps, 1) == [1., 0.]
-      @test coordinates(ps, 2) == [0., 1.]
+    
+    @test ps1.coords == ps2.coords
 
-      @test sprint(show, ps) == "2 PointSet{Float64,2}"
-      @test sprint(show, MIME"text/plain"(), ps) == "2 PointSet{Float64,2}\n 1.0  0.0\n 0.0  1.0"
+    # building a PointSet of tuples with different lengths should fail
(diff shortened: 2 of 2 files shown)


src/domains/point_set.jl
```diff
@@ -5,10 +5,12 @@
...
-A set of points with coordinate matrix `coords`. The number of rows
-of the matrix is the dimensionality of the domain whereas the number
-of columns is the number of points in the set. Alternatively, `coords`
-can be a vector of tuples (i.e. points).
+A set of points with coordinates vector `coords`. The vector
+contains many coordinates (number of points), where each coordinate
+is implemented by a static array of length same as the dimensionality.
+To build the PointSet a Matrix can be use. The matrix is the dimensionality 
+of the domain whereas the number of columns is the number of points in the set
+or a Vector of tuples that will be converted to a Vector of static array.
```
juliohm, line 13: The docstring can be more concise:

>A set of points with coordinates `coords`. Each point is represented by a static vector
>or tuple. Alternatively, `coords` can be a matrix where the number of rows equals the
> number of dimensions.

src/domains/point_set.jl
```diff
@@ -25,34 +27,44 @@ julia> PointSet([(rand(),rand()) for i in 1:100])
...
+function PointSet(coords::AbstractVector{NTuple{N,T}}) where {N,T}
+  points = Vector{SVector{N,T}}()
+  for row in coords
+    push!(points,SVector{N,T}(row))
+  end
+  PointSet{T,N}(points)
+end
 
-nelms(ps::PointSet) = size(ps.coords, 2)
+nelms(ps::PointSet) = size(ps.coords,1)
```
juliohm, line 55: This can be `length(ps.coords)`

src/domains/point_set.jl
```diff
@@ -25,34 +27,44 @@ julia> PointSet([(rand(),rand()) for i in 1:100])
...
+    buff[i] = ps.coords[location][i]
   end
 end
-
 # ------------
 # IO methods
 # ------------
 function Base.show(io::IO, ps::PointSet{T,N}) where {N,T}
-  npts = size(ps.coords, 2)
+  npts = nelms(ps)
```
juliohm, line 67: Let's use `length(ps.coords)` here as well just to avoid introducing dependencies between different function implementations.

src/domains/point_set.jl
```diff
@@ -25,34 +27,44 @@ julia> PointSet([(rand(),rand()) for i in 1:100])
 ```
 """
 struct PointSet{T,N} <: SpatialDomain{T,N}
-  coords::Matrix{T}
+  coords::Vector{SVector{N,T}} 
 
   function PointSet{T,N}(coords) where {N,T}
     @assert !isempty(coords) "coordinates must be non-empty"
     new(coords)
   end
```
juliohm, line 35: I think we can safely remove this inner constructor, what do you think? I don't remember why I introduced it, but it feels like boilerplate.

src/domains/point_set.jl
```diff
@@ -25,34 +27,44 @@ julia> PointSet([(rand(),rand()) for i in 1:100])
...
+end
 
-PointSet(coords::AbstractVector{NTuple{N,T}}) where {N,T} =
-  PointSet([c[i] for i in 1:N, c in coords])
+function PointSet(coords::AbstractVector{NTuple{N,T}}) where {N,T}
+  points = Vector{SVector{N,T}}()
+  for row in coords
+    push!(points,SVector{N,T}(row))
+  end
+  PointSet{T,N}(points)
```
juliohm, line 52: After you remove the inner constructor, perhaps this outer constructor won't be needed. In case it is needed, please use the following:

```julia
PointSet(coords::AbstractVector{<:NTuple}) = PointSet(SVector.(coords))
```

src/domains/point_set.jl
```diff
@@ -25,34 +27,44 @@ julia> PointSet([(rand(),rand()) for i in 1:100])
...
+end
 
-nelms(ps::PointSet) = size(ps.coords, 2)
+nelms(ps::PointSet) = size(ps.coords,1)
 
 function coordinates!(buff::AbstractVector{T}, ps::PointSet{T,N},
                       location::Int) where {N,T}
```
juliohm, line 58: Could you also rename this variable `location` to `ind`?

src/domains/point_set.jl
```diff
@@ -25,34 +27,44 @@ julia> PointSet([(rand(),rand()) for i in 1:100])
...
-PointSet(coords::AbstractMatrix{T}) where {T} =
-  PointSet{T,size(coords,1)}(coords)
+function PointSet(coords::AbstractMatrix{T}) where {T}
+  N = size(coords, 1)
+  points = Vector{SVector{N,T}}()
+  for row in eachcol(coords)
+    push!(points,SVector{N,T}(row))
+  end
+  PointSet{T,N}(points)
+end
```
juliohm, line 45: Perhaps this can be simplified to

```julia
PointSet(coords::AbstractMatrix) = PointSet(collect(eachcol(coords)))
```

This constructor could also come after the constructor with tuples.

src/domains/point_set.jl
```diff
@@ -25,34 +27,44 @@ julia> PointSet([(rand(),rand()) for i in 1:100])
...
 end
-
 # ------------
 # IO methods
 # ------------
 function Base.show(io::IO, ps::PointSet{T,N}) where {N,T}
-  npts = size(ps.coords, 2)
+  npts = nelms(ps)
   print(io, "$npts PointSet{$T,$N}")
 end
```
juliohm, line 70: Perhaps we should convert back to matrix before printing. Here is a one-liner:

```julia
X = reduce(hcat, ps.coords)
Base.print_array(io, X)
```

test/domains.jl
```diff
@@ -6,7 +6,7 @@
     @test coordinates(ps, 2) == [0., 1.]
 
     @test sprint(show, ps) == "2 PointSet{Float64,2}"
-    @test sprint(show, MIME"text/plain"(), ps) == "2 PointSet{Float64,2}\n 1.0  0.0\n 0.0  1.0"
+    @test sprint(show, MIME"text/plain"(), ps) == "2 PointSet{Float64,2}\n [1.0, 0.0]\n [0.0, 1.0]"
```
juliohm, line 9: We shouldn't need to change this one after we fix the `Base.show` method above.

src/domains/point_set.jl
exepulveda: But a point is indeed represented by a static vector, not a tuple. It can be initialised by a tuple.

src/domains/point_set.jl
exepulveda: Yes, It can.

src/domains/point_set.jl
exepulveda: OK

src/domains/point_set.jl
exepulveda: I agree

src/domains/point_set.jl
exepulveda: Much better indeed

src/domains/point_set.jl
exepulveda: Why is that? It seems a change over the whole code I guess to avoid confusion between location (coordinate) and ind (index). Please confirm. I have seen using location in many places.

src/domains/point_set.jl
exepulveda: Much better. I had that feeling that I short version would be possible, but not yet proficient with Julia :-)

src/domains/point_set.jl
exepulveda: I guess you want to preserve the way to output a point.

src/domains/point_set.jl
exepulveda: I like more:

> 
> A set of points with coordinates `coords`. Each point is internally represented by a static
> array. To build the PointSet a vector of tuples or a matrix can be use. 
> The matrix must be of rank 2, where the number of columns corresponds to
> the number of points and the number of rows is the dimensionality of each point

src/domains/point_set.jl
exepulveda: I need to inject the size of the point (N):

PointSet(coords::AbstractVector{<:NTuple}) = PointSet(SVector{N}.(coords))

Do you know a trick to do so without writing a function with two lines?

exepulveda: It seems if we have:
```
function PointSet(coords::AbstractVector{<:AbstractVector})
  @assert !isempty(coords) "coordinates must be non-empty"
  @assert length(coords) > 0 "coordinates must contain one point at least"
  N = length(coords[1]) #getting 
  PointSet(SVector{N}.(coords))
end
```
Then
```
PointSet(coords::AbstractMatrix) = PointSet(collect(eachcol(coords)))
PointSet(coords::AbstractVector{<:NTuple}) = PointSet(collect(map(x -> Vector(collect(x)),coords)))
```
works well

src/domains/point_set.jl
juliohm: Your suggestion is too verbose with a lot internal detail. A docstring is intended for end users who just need to know that they should provide a list of vectors or tuples. Can you please use the docstring I suggested above? It can even be made clearer if you replace `set of points` by `list of points`. Thanks.

src/domains/point_set.jl
juliohm: Thanks, looking forward to the change.

src/domains/point_set.jl
juliohm: I don't understand why the original version `PointSet(SVector.(coords))` didn't do the job. Why we need all these lines of code in the constructor above?

src/domains/point_set.jl
juliohm: It is not needed, but every occurrence of `location` is being replaced in the codebase by `ind` anyways. At some point they will go away. I thought it would be simple to do it here already.

src/domains/point_set.jl
juliohm: Yes, it is important to not break things when possible. The display is also nicer without a bunch of brackets hanging around.

src/domains/point_set.jl
exepulveda: Simply because it does not compile!
SVector needs somehow the size of the static array (N). SVector{N}.(coords)

src/domains/point_set.jl
exepulveda: I just was trying to keep as closer as possible to the doc before. But I can change it to what you propose.

src/domains/point_set.jl
exepulveda: I will try to make it working, but the suggested code did not make any change.

src/domains/point_set.jl
juliohm: Strange, it works for me:

```julia
julia> v = [(1.,2.),(3.,4.)]
2-element Array{Tuple{Float64,Float64},1}:
 (1.0, 2.0)
 (3.0, 4.0)
julia> SVector.(v)
2-element Array{SArray{Tuple{2},Float64,1,2},1}:
 [1.0, 2.0]
 [3.0, 4.0]
```
Let me check with the full signature...

juliohm: Perhaps this is about the Julia version? I am on Julia v1.5

src/domains/point_set.jl
juliohm: In the past we used `location` to mean the integer index, it was always equivalent to `ind`. Nowadays it is just clearer to use `ind` everywhere because sometimes this index is used in the table of values as opposed to in the spatial domain.

src/domains/point_set.jl
juliohm: The variable `X` should be a matrix in the snippet of code above assuming that `ps.coords` is already in the right format as a vector of static vector.

src/domains/point_set.jl
exepulveda: I'm using v1.5.1

exepulveda: It works for that case, but not in this case:
```
a = [[1.0, 2.0],[10.,20.]]
SVector.(a)
```

juliohm: Yes, we don't want to support `AbstractVector{<:AbstractVector}`. Tuples and static vectors have known length at compile time, unlike standard vectors.

At the end we should have 3 constructors:
- Inner constructor for list of SVector which is created by the compiler
- Outer constructor for list of Tuple which calls the inner constructor after a `SVector.(v)` conversion
- Outer constructor for 2D matrix which uses the number of rows as `N` in `SVector{N}`

exepulveda: I think that is lack of flexibility for uses. Using vector of vector is quite common.
Interestingly, if you use a matrix and iterate the columns, you get a Vector (View) for each column not a tuple.
```

julia> a = rand(2,4)
2×4 Array{Float64,2}:
 0.409933  0.427235  0.487495  0.327178
 0.316636  0.845672  0.424418  0.581679

julia> first(eachcol(a))
2-element view(::Array{Float64,2}, :, 1) with eltype Float64:
 0.40993283206458475
 0.3166363584400025
```

juliohm:
Yes but we don't want to incentivate the construction of vector of vectors.
We only keep the constructor with a matrix because it is convenient for
quick workflows. This is a standard across the Julia geo ecosystem.

src/domains/point_set.jl
```diff
@@ -3,13 +3,11 @@
 # ------------------------------------------------------------------
 
 """
-    PointSet(coords)
-
-A set of points with coordinate matrix `coords`. The number of rows
-of the matrix is the dimensionality of the domain whereas the number
-of columns is the number of points in the set. Alternatively, `coords`
-can be a vector of tuples (i.e. points).
+PointSet(coords)
```
juliohm, line 6: Can you please format the docstring as before? 4 whitespaces before the `PointSet` title, and backquotes for the `coords`. Also, an extra space before the header `## Examples`.

src/domains/point_set.jl
```diff
@@ -25,38 +23,30 @@ julia> PointSet([(rand(),rand()) for i in 1:100])
...
 function Base.show(io::IO, ps::PointSet{T,N}) where {N,T}
-  npts = size(ps.coords, 2)
+  npts = length(ps.coords)
   print(io, "$npts PointSet{$T,$N}")
 end
 
 function Base.show(io::IO, ::MIME"text/plain", ps::PointSet{T,N}) where {N,T}
   println(io, ps)
-  Base.print_array(io, ps.coords)
+  m = hcat(ps.coords...)
```
juliohm, line 50: Please use `reduce(hcat, ps.coords)`. Splatting large collections with `...` is not efficient.

test/domains.jl
```diff
@@ -1,21 +1,25 @@
 @testset "Domains" begin
   @testset "PointSet" begin
-    ps = PointSet([1. 0.; 0. 1.])
-    @test nelms(ps) == 2
-    @test coordinates(ps, 1) == [1., 0.]
-    @test coordinates(ps, 2) == [0., 1.]
```
juliohm, line 6: Can you please undo these changes and add a separate section for the new constructor? We just need a small paragraph constructing a point set with a list of tuples and doing basic checks.

juliohm (changes requested):
Thank you for the updates. Attached a new set of comments.

test/domains.jl
exepulveda: We will be duplicating code. I don't like to duplicate code.

juliohm:
You have to understand that the change that you're making is minimal to
these constructors. It shouldn't affect the test suite so dramatically.
Only a few extra lines using the new constructor would do it. Even no extra
tests would be fine since this change is only about internal
representation. Also please be considerate in your comments, they sound a
bit rude sometimes. I know it is not your intention.

exepulveda:
Of course is not my intention. I had the same feeling about some of your comments and I knew that it was not your intention, they are just manners and we need to learn how to work together with respect. Happy to talk you about this if that helps and help me to realise what comments were a bit rude to you.

Coming back to the point, we are here to collaborate. 

If you think that the test that I'm changing for the good is not reasonable, I need to understand the reasons. Testing both ways to build a PointSet without duplicating code is a good thing. The reason is simple: independent of the way the PointSet was built, all the following tests should pass. Am I wrong?

juliohm:
Thanks. Please let me know which comments I made that sounded rude so that I can't repeat them again. Written text is always tricky, and depending on where you write from you can sound differently to others. None of us is a native English speaker. So keeping this in mind is always a good idea. Assume good intentions :+1: 

I agree that tests should cover both constructors as much as possible. The fact is we already had both constructors before, we only changed the internal behavior. We can add more tests because that never hurts, but we should do it in a way that doesn't refactor the entire test set for the point set. Particularly, you wrapped the visual regression tests in a loop without modifying the file names so two figures were generated on the same file, one overwriting the other. Also, checking for the `show` methods twice doesn't help much. That is why I thought that it would be much simpler to just add a separate section with a few more simple tests for the other constructor. Makes sense?

exepulveda:
We can talk about what things seem a bit rude later face-to-face :-)

We are not refactoring the entire test. I just want to make sure that for the same values, but different representations, the PointSet are the same. I will make some adjustments trying to keep as much as possible the existing tests code.

src/domains/point_set.jl
```diff
@@ -25,38 +24,30 @@ julia> PointSet([(rand(),rand()) for i in 1:100])
...
 function Base.show(io::IO, ps::PointSet{T,N}) where {N,T}
-  npts = size(ps.coords, 2)
+  npts = length(ps.coords)
   print(io, "$npts PointSet{$T,$N}")
 end
 
 function Base.show(io::IO, ::MIME"text/plain", ps::PointSet{T,N}) where {N,T}
   println(io, ps)
-  Base.print_array(io, ps.coords)
+  m = reduce(hcat, ps.coords) #hcat(ps.coords...)
```
juliohm, line 51: can you please remote the comment after the line?

test/domains.jl
```diff
@@ -1,13 +1,25 @@
 @testset "Domains" begin
   @testset "PointSet" begin
+    # testing different ways to build a PointSet results in the same PointSet
+    ps1 = PointSet([1. 0.; 0. 1.])            # as matrix
+    ps2 = PointSet([(1.0, 0.0), (0.0, 1.0)])  # as vector of tuples
+    
+    @test ps1.coords == ps2.coords
+
+    # building a PointSet of tuples with different lengths should fail
+    @test_throws MethodError PointSet([(1.0, 0.0), (0.0, 1.0 , 4.0)])
```
juliohm, line 10: Nice! Much easier to review these changes.

juliohm (changes requested):

src/domains/point_set.jl
exepulveda: Yeap. I forgot to do it before committing.

src/domains/point_set.jl
exepulveda: > Yes, we don't want to support `AbstractVector{<:AbstractVector}`. Tuples and static vectors have known length at compile time, unlike standard vectors.
> 
> At the end we should have 3 constructors:
> 
>     * Inner constructor for list of SVector which is created by the compiler
> 
>     * Outer constructor for list of Tuple which calls the inner constructor after a `SVector.(v)` conversion
> 
>     * Outer constructor for 2D matrix which uses the number of rows as `N` in `SVector{N}`

Vector of vector and Vector of tuple are quite the same from user perspective. The important thing is we will normilised them to an internal representation that is much more performant. Also, both representations may fail if the length of each point are different. The current implementation assumes that the final point length will be equal to the first point length. Matrix does not have that issue.

juliohm:
It is possible to restrict the length of the tuples to be the same in the
type signature. NTuple{N,T} if I'm not mistaken. Typing from my phone...
calling the day in BRT time here.

juliohm:
I will squash, merge and fix the last details on master so that this change is included in the next release :+1: Thanks again for the improvements, always nice to see speedups :)

PR #54, opened 2020-09-28
Bump compat for Optim to 1.2

ElOceanografo:


juliohm:
Thanks. It is strange that CompatHelper didn't trigger the update. I will wait for the tests and then merge 👍

PR #55, opened 2020-10-01
Flexibility to rotation conventions

rmcaixeta:
Starting the PR for you to check out if it seems ok

I was adapting the previous structure to the initial idea. But realized that ReferenceFrameRotations.jl has almost all of that plus quaternions, so I followed that way. If that's ok, I guess I can finish all the tests in the next few days, maybe there's still adjustments in some of the convention rules.

GSLIB is matching perfectly now:

```julia
ang1,ang2,ang3 = [30.,30.,30.]
a_hmax,a_hmin,a_vert = [50.,25.,5.]
rot = Ellipsoidal([a_hmax,a_hmin,a_vert], [ang1,ang2,ang3], convention=:GSLIB)
γ = SphericalVariogram(sill=1.0, nugget=0.1, distance=rot)
```

Tried several rotations and they're all matching vmodel output. @exepulveda , I guess that with the current Ellipsoidal function you might have problems to match complex rotation with GSLIB if you're trying that (e.g 30, 30, 30). One uses intrinsic rotations and the other uses extrinsic, it may be quite hard to match them if you rotate more than one axis

Finally, one question about quaternions. Semiaxes scaling vector is somehow stored into the quaternions too or only the rotation? That was not totally clear for me when I made the `Ellipsoidal(semiaxes,quaternion)` function.


Project.toml
```diff
@@ -50,6 +51,7 @@ StaticArrays = "0.12"
 StatsBase = "0.32, 0.33"
 Tables = "1.0"
 Transducers = "0.4"
+ReferenceFrameRotations = "0.5"
```
juliohm, line 54: Amazing. I was just considering ReferenceFrameRotations.jl today. The package is well-written and well-documented, so we shouldn't have problems adding it as a dependency :+1:

src/GeoStatsBase.jl
```diff
@@ -26,6 +26,7 @@ using ScientificTypes
 using LossFunctions
 using RecipesBase
 using Parameters
+using ReferenceFrameRotations: angle_to_dcm, quat_to_dcm, Quaternion
```
juliohm, line 29: Mind moving this line to right after `using NearestNeighbors`? Also, if you can move the `Quaternion` to the beginning before the function names, that would be great.

src/distances.jl
```diff
@@ -3,13 +3,21 @@
...
 
 """
-    Ellipsoidal(semiaxes, angles)
+    Ellipsoidal(semiaxes, angles; convention=:TaitBryanExtr)
 
 A distance defined by an ellipsoid with given `semiaxes` and rotation `angles`.
+Default angle convention is Tait-Bryan ZXY, extrinsic right-handed rotation
+Other conventions available: :TaitBryanIntr, :EulerExtr, :EulerIntr,
+                             :GSLIB, :Leapfrog, :Datamine
+* :EulerExtr, :EulerIntr and :Datamine assumes ZXZ rotation sequence
```
juliohm, line 12: Could you create a section in the docstring with `## Conventions` and then add each convention as a bullet? It is awesome that you added support to so many conventions, including those from popular commercial software :100:

src/distances.jl
```diff
@@ -3,13 +3,21 @@
...
 
 - For 2D ellipsoids, there are two semiaxes and one rotation angle.
 - For 3D ellipsoids, there are three semiaxes and three rotation angles.
 
+    Ellipsoidal(semiaxes, quaternion)
+
+Alternatively, give the `semiaxes` and the orientation as a `quaternion`
+
```
juliohm, line 20: I am also curious if we should even include this quaternion constructor. It seems that the other constructor will be super useful already and will cover all cases that one could cover with quaternions, agree? What do you think?

src/distances.jl
```diff
@@ -24,53 +32,76 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
 julia> Ellipsoidal([1.0,0.5,0.5], [π/2,0.0,0.0])
 ```
 """
+
+# Right-hand rule; motion defined looking to the negative direction of the axis
+struct RotationRules
```
juliohm, line 37: Perhaps rename the struct to singular `RotationRule` given that it only stores a single rule.

src/distances.jl
```diff
@@ -24,53 +32,76 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
...
+  rot_seq::Symbol
```
juliohm, line 38: Maybe rename this to `order` or `sequence`?

src/distances.jl
```diff
@@ -24,53 +32,76 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
...
+  motion::Vector{Symbol}
+  radian::Bool
+  main::Symbol
+  extrinsic::Bool
+end
+
+themes = Dict(
```
juliohm, line 45: Maybe rename this variable to `conventions`

src/distances.jl
```diff
@@ -24,53 +32,76 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
...
+  :TaitBryanExtr => RotationRules(:ZXY,[:CCW,:CCW,:CCW],true,:x,true),
+  :TaitBryanIntr => RotationRules(:ZXY,[:CCW,:CCW,:CCW],true,:x,false),
+  :EulerExtr => RotationRules(:ZXZ,[:CCW,:CCW,:CCW],true,:x,true),
+  :EulerIntr => RotationRules(:ZXZ,[:CCW,:CCW,:CCW],true,:x,false),
+  :GSLIB => RotationRules(:ZXY,[:CW,:CCW,:CCW],false,:y,false),
+  :Leapfrog => RotationRules(:ZYZ,[:CW,:CW,:CW],false,:x,false),
+  :Datamine => RotationRules(:ZXZ,[:CW,:CW,:CW],false,:x,false)
```
juliohm, line 52: It would be nice to align the arrows of the dict for readability.

src/distances.jl
juliohm: Or maybe `rules` so that when we access the key in the constructor it reads like `rules[convention]`.

src/distances.jl
```diff
@@ -24,53 +32,76 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
...
   dist::Mahalanobis{T}
 
-  function Ellipsoidal{N,T}(semiaxes, angles) where {N,T}
+  function Ellipsoidal{N,T}(semiaxes, angles; convention=:TaitBryanExtr) where {N,T}
     @assert length(semiaxes) == N "number of semiaxes must match spatial dimension"
     @assert all(semiaxes .> zero(T)) "semiaxes must be positive"
     @assert N ∈ [2,3] "dimension must be either 2 or 3"
 
+    # invert x and y if necessary
+    if themes[convention].main == :y
```
juliohm, line 64: I would first create a rule object with `rule = rules[convention]`, possibly checking first if the provided convention is a key of the dictionary and throwing an error otherwise. Then, we could use the rule object in the rest of the code like `rule.main`, `rule.sequence`, ...

src/distances.jl
```diff
@@ -24,53 +32,76 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
...
     # rotation matrix
     if N == 2
-      θ = angles[1]
-
-      cosθ = cos(θ)
-      sinθ = sin(θ)
-
-      P = [cosθ -sinθ
-           sinθ  cosθ]
+      θ = themes[convention].motion[1] == :CCW ? angles[1] : -1*angles[1]
```
juliohm, line 76: I think we can just do `-angles[1]`

src/distances.jl
```diff
@@ -24,53 +32,76 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
...
+      P = angle_to_dcm(θ, 0, 0, themes[convention].rot_seq)[1:2,1:2]
+
+    else # N == 3
+      for (i,sign) in enumerate(themes[convention].motion)
+        angles[i] = sign == :CCW ? angles[i] : -1*angles[i]
```
juliohm, line 81: same here with `-angles[i]`

src/distances.jl
```diff
@@ -24,53 +32,76 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
 julia> Ellipsoidal([1.0,0.5,0.5], [π/2,0.0,0.0])
 ```
 """
+
+# Right-hand rule; motion defined looking to the negative direction of the axis
```
juliohm, line 36: Perhaps it would be a good idea to add a more detailed docstring for this struct explaining the meaning of each of the fields. That can help people interested in adding new conventions, though I thin you already covered it all :)

src/distances.jl
```diff
@@ -24,53 +32,76 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
...
-      Rxy = [cosxy -sinxy _0
-             sinxy  cosxy _0
-             _0     _0 _1]
+    new(Mahalanobis(Q))
+  end
 
-      Ryz = [_1    _0     _0
-             _0 cosyz -sinyz
-             _0 sinyz  cosyz]
+  function Ellipsoidal{N,T}(semiaxes, quaternion::Quaternion) where {N,T}
```
juliohm, line 92: I vote for just removing this constructor for now, unless you are seeing a good reason to keep it in practice. My intuition says that not many people would use it.

juliohm (changes requested):
This is an amazing PR @rmcaixeta :heart: Thank you for putting the effort, and for the careful code style. I've added a few comments and suggestion for us to start reviewing it together.

rmcaixeta:
Thanks for the invite and the suggestions. I'll work on your points here

juliohm:
You are welcome, it is good to have you here :) Could you please make your membership public? GitHub makes it private by default for some reason.

Project.toml
rmcaixeta: Nice. It's a really nice library for rotations

src/GeoStatsBase.jl
rmcaixeta: Done. Removed quaternion functions (discussion in next commments)

src/distances.jl
rmcaixeta: Done, see if it's okay

src/distances.jl
rmcaixeta: I'm not used to work with quaternions either. Thought about putting it in case some user prefer working with them. But I agree with you. In case someone is working with quaternions, it's possible to easily convert it to angles via ReferenceFrameRotations.jl `quat_to_angles()`

src/distances.jl
rmcaixeta: I guess the only cases not covered are strange rotations such as :YZX or something like that. About that. Do you think it'd be advisable to export RotationRule() and give the option to inform custom rules or let restrict it inside the code?

src/distances.jl
rmcaixeta: Done

src/distances.jl
rmcaixeta: Done

src/distances.jl
rmcaixeta: Done

src/distances.jl
rmcaixeta: Done

src/distances.jl
rmcaixeta: Nice. Done it

src/distances.jl
rmcaixeta: Check out what you think about the change

rmcaixeta:
Enough for today. I'll prepare some formal tests tomorrow

src/distances.jl
juliohm: Let's keep it inside the code for now, I think most people won't bother creating their own rules, and will instead pick one of the available ones.

src/distances.jl
```diff
@@ -24,64 +84,50 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
...
+      for (i,sign) in enumerate(rule.motion)
+        invert = (sign==:CW && !rule.extrinsic) || (sign==:CCW && rule.extrinsic)
+        angles[i] = invert ? -angles[i] : angles[i]
+      end
+      P = angle_to_dcm(angles[1], angles[2], angles[3], rule.order)
     end
 
     # ellipsoid matrix
-    Q = P*Λ*P'
+    Q = rule.extrinsic ? P*Λ*P' : P'*Λ*P
```
juliohm, line 123: Just to confirm, this is still needed even after the sign flip above? Thanks for the careful PR, these angles are super tricky to get right.

juliohm (changes requested):
Nice round of review. Attached a few more comments and questions. After that I think we just need the tests before we can merge :100:

src/distances.jl
juliohm: Also, it would be nice to double check the the resulting Q matrix is static after the multiplication with the diagonal.

src/distances.jl
```diff
@@ -24,64 +84,50 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
...
+      invert = (rule.motion[1]==:CW && !rule.extrinsic) || (rule.motion[1]==:CCW && rule.extrinsic)
+      θ = invert ? -angles[1] : angles[1]
+      P = angle_to_dcm(θ, 0.0, 0.0, rule.order)[1:2,1:2]
+    else # N == 3
+      for (i,sign) in enumerate(rule.motion)
+        invert = (sign==:CW && !rule.extrinsic) || (sign==:CCW && rule.extrinsic)
+        angles[i] = invert ? -angles[i] : angles[i]
+      end
+      P = angle_to_dcm(angles[1], angles[2], angles[3], rule.order)
     end
```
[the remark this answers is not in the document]
rmcaixeta: Nice, much more compact now

src/distances.jl
rmcaixeta: This sign flip is still necessary indeed
Checked about static matrix. Both P and Λ were as arrays (the first after reindexing it [1:N,1:N])
But after "force" StaticArrays, I got that:

```
ERROR: LoadError: MethodError: no method matching Distances.Mahalanobis(::StaticArrays.SArray{Tuple{2,2},Float64,2,4})
Closest candidates are:
  Distances.Mahalanobis(::Array{T,2}) where T
```

Seems to be some restriction in the `Distances.jl` (??)

juliohm: It seems that Mahalanobis and SqMahalanobis could be improved: https://github.com/JuliaStats/Distances.jl/blob/d547d3a78db98411a902fe5382243ece00bfdcc3/src/mahalanobis.jl#L3-L9

Do you mind submitting a PR to Distances.jl to add a parameter type like:

```julia
struct Mahalanobis{T,M<:AbstractMatrix{T}} <: Metric
    qmat::M
end
```
or even the cleaner:

```julia
struct Mahalanobis{M<:AbstractMatrix} <: Metric
    qmat::M
end
```

If they accept the improvement, we can work with static matrices, which should be much master. Otherwise, we will need to work a custom Mahalanobis in our side.

rmcaixeta: Yeah
Well, I confess that these Julia parametric types/methods are still quite confusing for me.
But ok, I'll copy and paste your code in a issue, let's see how it goes

rmcaixeta:
https://github.com/JuliaStats/Distances.jl/pull/182

Let's wait for it
Meanwhile, it's working good with standard arrays

juliohm:
I've commented on the issue there. We need to guarantee that the type is concrete.

src/distances.jl
```diff
@@ -24,64 +84,47 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
...
+    N == 2 && (angles = [angles[1], _0, _0])
+    intr = @. (rule.motion == :CW) & !rule.extrinsic
+    extr = @. (rule.motion == :CCW) & rule.extrinsic
+    angles[intr .| extr] *= -1
 
-      P = Rzx*Ryz*Rxy
-    end
+    # rotation matrix
+    P = angle_to_dcm(angles..., rule.order)[1:N,1:N]
+    #P = SMatrix{N,N}(angle_to_dcm(angles..., rule.order)[1:N,1:N])
```
juliohm, line 117: I think angle_to_dcm already returns a static matrix?

src/distances.jl
```diff
@@ -24,64 +84,47 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
...
-      Rxy = [cosxy -sinxy _0
-             sinxy  cosxy _0
-             _0     _0 _1]
-
-      Ryz = [_1    _0     _0
-             _0 cosyz -sinyz
-             _0 sinyz  cosyz]
+    # scaling matrix
+    Λ = Diagonal(one(T)./semiaxes.^2)
+    #Λ = SMatrix{N,N}(Diagonal(one(T)./semiaxes.^2))
```
juliohm, line 105: I think we can just `SMatrix(Diagonal(...))`, we don't need the `{N,N}` there.

rmcaixeta:
> I've commented on the issue there. We need to guarantee that the type is concrete.

I made the simple fix because with your suggestion it raised some other stuff:

```
ERROR: LoadError: MethodError: Cannot `convert` an object of type 
  Distances.Mahalanobis{StaticArrays.SArray{Tuple{2,2},Float64,2,4}} to an object of type 
  Union{}
```

src/distances.jl
rmcaixeta: Yes. But when I do [1:N,1:N] it seems to convert back to `Matrix`

juliohm:
Yes, I think the PR needs to update the other functions in the mahalanobis.jl file to replace the type parameter `T` by the new type parameters `T, M`.

For example, assume we have this new type:

```julia
struct Mahalanobis{T,M<:AbstractMatrix{T}} <: Metric
    qmat::M
end
```

We need to rewrite the function:

```julia
function (dist::SqMahalanobis{T})(a::AbstractVector, b::AbstractVector) where {T <: Real}
    if length(a) != length(b)
        throw(DimensionMismatch("first array has length $(length(a)) which does not match the length of the second, $(length(b))."))
    end

    Q = dist.qmat
    z = a - b
    return dot(z, Q * z)
end
```

to

```julia
function (dist::SqMahalanobis{T,M})(a::AbstractVector, b::AbstractVector) where {T,M}
    if length(a) != length(b)
        throw(DimensionMismatch("first array has length $(length(a)) which does not match the length of the second, $(length(b))."))
    end

    Q = dist.qmat
    z = a - b
    return dot(z, Q * z)
end
```

And do similar fixes for the other functions that assumed a single type parameter `T`. Makes sense?

src/distances.jl
juliohm: Wow, that is surprising. I wonder if we should ask it on Zulip. This behavior is certainly strange.

src/distances.jl
juliohm: Nevermind, we actually need the sizes explicitly with matrices.

rmcaixeta:
Got it. I'll try that

rmcaixeta:
Is that change correct?

```
result_type(::Mahalanobis{T,M}, ::Type, ::Type) where {T,M} = T
```

juliohm:
Yes :+1: The line is saying that the result type of the distance is the same type of the entries of the matrix.

juliohm:
Also, in the snippet I shared above, you could also eliminate the type parameters completely as they are not used in the body:

```julia
function (dist::SqMahalanobis)(a::AbstractVector, b::AbstractVector)
    if length(a) != length(b)
        throw(DimensionMismatch("first array has length $(length(a)) which does not match the length of the second, $(length(b))."))
    end

    Q = dist.qmat
    z = a - b
    return dot(z, Q * z)
end
```

juliohm:
In general, you have the following:

```julia
julia> struct Foo{T} end

julia> Foo{Float64} <: Foo{T} where T # true because Float64 is a specific instance of T
true

julia> (Foo{T} where T) <: Foo # true because Foo{T} is a subtype of the generic Foo
true
```

rmcaixeta:
Thanks for the explanation. I'm much more a 'scripter' than a 'programmer', these structure definitions are still a little confuse for me

juliohm:
No worries, it takes some time to get used to these parametric types. If you have experience in C++, they are similar to template parameters, but much simpler to work with.

In general, we always try to make fields of a struct of a concrete type for performance. When the field type is trivial like `Int` or `Vector{Float64}` etc, we can just move on with type parameters:

```julia
struct Foo
  a::Int
  b::Vector{Float64}
end
```

However, sometimes we want to allow types that aren't available yet, or types that are a subtype of a broader collection like all matrices in the language that subtype from `AbstractMatrix`:

```julia
struct Bar{M<:AbstractMatrix}
  a::M
end
```

We could have left the code as:

```julia
struct Bar
  a
end
```
but then the type would be `Any` which is also slow as the generic abstract types. For some structs we don't need to bother about the field type, because function calls correct the types for us. For other structs that are used over and over in tight loops for example, we need to be careful to help the compiler from the beginning.

src/distances.jl
juliohm: Currently we have to do `[SOneTo(N),SOneTo(N)]` to get a static array back. Could you adapt it here?

Source: https://discourse.julialang.org/t/recent-developments-in-staticarrays-jl-land/47778/6?u=juliohm

src/distances.jl
juliohm: I think this one can be `SDiagonal(ntuple(i->one(T)/semiaxes[i]^2, N))`

rmcaixeta: Done

src/distances.jl
rmcaixeta: Done. All working locally. Just need to wait the new version of Distances.jl to work properly for all

exepulveda:
An additional point we need to address soon with rotations is that for searches, KDTree only support the following metrics: Euclidean, Chebyshev, Cityblock, Minkowski, WeightedEuclidean, WeightedCityblock, WeightedMinkowski (see https://github.com/KristofferC/NearestNeighbors.jl/blob/master/src/NearestNeighbors.jl line 24)

Our neigbour searches are based on KDTree and they would not work with custom distances.

juliohm:
Regarding the KDTree, I think we could do something similar to what we currently do in the inverse distance weighting solver here: https://github.com/JuliaEarth/GeoEstimation.jl/blob/4135ad518a97ba128bd17c86a77d1d337caf59ae/src/idw.jl#L57-L61

Basically, we check if the distance is one of the permitted ones for KDTree and fallback to brute force. This could be done in a separate PR for the searcher methods.

exepulveda:
> Regarding the KDTree, I think we could do something similar to what we currently do in the inverse distance weighting solver here: https://github.com/JuliaEarth/GeoEstimation.jl/blob/4135ad518a97ba128bd17c86a77d1d337caf59ae/src/idw.jl#L57-L61
> 
> Basically, we check if the distance is one of the permitted ones for KDTree and fallback to brute force. This could be done in a separate PR for the searcher methods.

We should do everything to avoid using BruteForce. One option is determining metrics based on rotations, and build the KDTree with rotated coordinates and search after rotating the query point. This should work. Does this make sense?

juliohm:
Makes sense @exepulveda. :+1: I think in order to achieve this, we will need to introduce a rotation API first for the objects in GeoStats.jl that consumes rotations as specified by CoordinateFrameRotations.jl for example and performs the rotations. One of the challenges right now is that there are many different packages with different transformations, e.g. CoordinatesTransformations.jl, Rotations.jl, CoordinatesFrameRotations.jl, etc. Each of them implementing a subset of transformations. We need to unify these transformations in a single place before we move forward.

rmcaixeta:
I guess it can be done with the `ReferenceFrameRotations.jl` used for this PR
Probably would be necessary to make the coordinate transformations at every point. Maybe the same matrix generated for `Ellipsoidal` could be used, but need to test it. In that case, I could split the `Ellipsoidal{N,T}(semiaxes, angles; convention=:TaitBryanExtr)` in two parts. One separate function `TransformationMatrix{N,T}(semiaxes, angles; convention=:TaitBryanExtr)` would retrieve the transformation matrix for both Mahalanobis and coordinate transformation. And the other function `Ellipsoidal{N,T}(Q)` would only use the matrix to create the Mahalanobis distance object

juliohm:
I was having the same thought the other day @rmcaixeta :+1: Perhaps we should split up the `Ellipsoidal` struct we have currently into two *functions*, one for constructing rotation matrices with the conventions you introduced, and possibly another that produces the Mahalanobis object. Should we address this proposal in a separate PR for organization purposes?

The tests are failing because NearestNeighbors.jl didn't update the version of Distances.jl yet: https://github.com/KristofferC/NearestNeighbors.jl/pull/111 After they merge and tag a new release, we can trigger the build here again to make sure the tests are passing.

rmcaixeta:
Great. It seems to be an easy fix, I think I can do that tomorrow

juliohm:
Nice. Let's do it in a separate PR after this one is merged. This PR here is already very useful in its current form.

src/distances.jl
```diff
@@ -24,64 +84,45 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
 julia> Ellipsoidal([1.0,0.5,0.5], [π/2,0.0,0.0])
 ```
 """
+
 struct Ellipsoidal{N,T} <: Metric
```
dkarrasch, line 88: As is, this contains an abstractly typed field. I think it should be
```julia
struct Ellipsoidal{N,M<:Mahalanobis} <: Metric
    dist::M
...
end
```

dkarrasch (commented):
I was just passing by to see what you guys are doing with the Mahalanobis thing. I commented on one detail. Hope it helps.

src/distances.jl
juliohm: Thanks for catching this @dkarrsch, very good point. @rmcaixeta could you fix it? This will also retrigger the build for us.

rmcaixeta: Thanks @dkarrasch. Updating here in some minutes

rmcaixeta:
> Nice. Let's do it in a separate PR after this one is merged. This PR here is already very useful in its current form.

Ok. Currently it's also possible to get the transformation matrix `Q` for eventual tests by that: 

```julia
d₁ = Ellipsoidal([1.,1.], [0.])
Q = d₁.dist.qmat
```

juliohm:
Yes, we should definitely cleanup this `Ellipsoidal` type, and make it two separate functions. This struct is too clumsy, and dates back to a very old version of the project. Now that we are starting to add support for rotations, we can refactor and get rid of it slowly.

src/distances.jl
```diff
@@ -24,64 +84,45 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
...
 ```
 """
-struct Ellipsoidal{N,T} <: Metric
-  dist::Mahalanobis{T}
 
-  function Ellipsoidal{N,T}(semiaxes, angles) where {N,T}
+struct Ellipsoidal{N,M<:Mahalanobis} <: Metric
+  dist::M
+
+  function Ellipsoidal{N,T}(semiaxes, angles; convention=:TaitBryanExtr) where {N,T}
```
juliohm, line 91: @rmcaixeta could you please replace the T by M here as well?

src/distances.jl
```diff
@@ -24,64 +84,45 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
...
+    Q = rule.extrinsic ? P*Λ*P' : P'*Λ*P
 
     new(Mahalanobis(Q))
   end
 end
 
-Ellipsoidal(semiaxes::AbstractVector{T}, angles::AbstractVector{T}) where {T} =
-  Ellipsoidal{length(semiaxes),T}(semiaxes, angles)
+Ellipsoidal(semiaxes::AbstractVector{T}, angles::AbstractVector{T}; kwargs...) where {T} =
+  Ellipsoidal{length(semiaxes),T}(semiaxes, angles; kwargs...)
```
juliohm, line 125: Also this constructor needs an update after the new type parameter M

src/distances.jl
```diff
@@ -24,64 +84,45 @@ julia> Ellipsoidal([1.0,0.5], [π/2])
...
 
 evaluate(dist::Ellipsoidal{N,T}, a::AbstractVector, b::AbstractVector) where {N,T} =
```
juliohm, line 127: T -> M

src/distances.jl
rmcaixeta: Would need some extra changes. `one(M)` and `zero(M)` does not work.  Is this ok?

```julia
function Ellipsoidal{N,M}(semiaxes, angles; convention=:TaitBryanExtr) where {N,M}
    T = eltype(semiaxes)
```

Or do I replace the following `one(T)` and `zero(T)` for `1.0` and `0.0`? Λ may not end as a matrix of Floats, but `P` will be StaticMatrix{Float64} (even `cos(zero(Int))` return Float64) and Q=P*Λ*P' will also be a float

src/distances.jl
rmcaixeta: Could make it work with some extra changes too. Is that ok?

```julia
Ellipsoidal(semiaxes::AbstractVector, angles::AbstractVector; kwargs...) =
  Ellipsoidal{length(semiaxes),Mahalanobis}(semiaxes, angles; kwargs...)

evaluate(dist::Ellipsoidal{N,M}, a::AbstractVector, b::AbstractVector) where {N,M} =
  evaluate(dist.dist, a, b)
```

juliohm:
Perhaps we should implement the changes you suggested before. What about we
first replace this Ellipsoid struct by a simple function that returns the
underlying Mahalanobis distance? Maybe ellipsedist would be a good name.
That way we get rid of these type parameters in the struct.

src/distances.jl
juliohm: Regarding the `cos` and `sin`, I think we should watch for the case with single precision `cos(1.0f0)`, that works correctly. The case with `Int` is less important in applications, and we can live with the return type `Float64`.

exepulveda:
Is Mahalanobis distance equivalent to Euclidian + Rotation?

juliohm:
Mahalanobis stretches the axes besides the rotation to create the
anisotropy.

exepulveda:
So Mahalanobis = Euclidean + Rotation + Scaling. I wonder if the scaling are equivalents. I will have a look at the maths :-)

juliohm:
@rmcaixeta just sharing in case you don't know this already: you can create PRs directly from a new branch in our repositories within JuliaEarth. That makes it easier for us to update and improve PRs together. Please let me know if you need some help with those last changes we discussed. I am looking forward to merging the nice work you've done here.

rmcaixeta:
Nice, I'll work with JuliaEarth branches for the next ones

I'll do the changes today after work
Meanwhile, if you could help about the compat issues. I think there are still some deps that does not accept Distances>=0.10, but not sure which ones.
Or I can do something ugly like `Distances==0.9 && convert(StaticArray,Array)` (don't know if that's possible) or just `convert(StaticArray,Array)` to be able to merge

juliohm:
I will check the compat issues during the day :ok_hand:

juliohm:
Opened a few PRs to fix the Distances.jl version:

- https://github.com/alan-turing-institute/MLJModels.jl/pull/317
- https://github.com/JuliaStats/Clustering.jl/pull/204

rmcaixeta:
Great, thanks!
Finishing it here, I'll update in some minutes.
About the function name, what do you think of **`anisotropy(args...)`** or `distance =`**`anisotropic(args...)`**?
As it's just an implementation of geometric anisotropy (though **`geometricanisotropy`** seems too big)

juliohm:
Nice. I was thinking if we could figure out a name that reminds us that this is also a distance, something like `aniso2distance`? What do you think? The more specific is the name, the easier it is to memorize the functionality. Perhaps `anisotropic` is too general, and doesn't ring a bell about it being a distance at the end of the day.

rmcaixeta:
Let's go with `aniso2distance` for now then. If we think of something better until all compat issues are closed, we change it later. Will this trigger changes to other packages? I remember there is at least one tutorial using `Ellipsoidal`

juliohm:
Thanks for the suggestion by the way. I was so used to "ellipsoid" that I didn't realize how bad that name actually was. Let's go with `aniso2distance` then :+1: We can revise any naming in the future, and the good thing about `aniso2distance` is that it won't clash with any other name we might need for other features.

Regarding the tutorials repository, we don't need to worry about it. It has its own Project.toml and Manifes.toml files so they won't get the updates unless we manually go there and update the environment with a new release of the stack :ok_hand:

juliohm (approved):

juliohm:
Thank you again for your work @rmcaixeta. This is a top-class PR that will certainly help a lot of people. Let me say also that I appreciate that you pay attention to minor details, it really helps in the review process. Looking forward to improving the project together.

Given that the issues have been fixed in the test dependencies, I will go ahead and merge it :ok_hand:

rmcaixeta:
Thanks man, happy to contribute to the project

PR #60, opened 2020-10-24
Cross validations

rmcaixeta:
Opening a pull request to organize the discussion to solve https://github.com/JuliaEarth/GeoStats.jl/issues/128

### DIFF
--- a/src/errors/block_cross_validation.jl
+++ b/src/errors/block_cross_validation.jl
@@ -30,11 +30,11 @@ BlockCrossValidation(sides; loss=Dict()) =
 function error(solver::AbstractSolver,
                problem::AbstractProblem,
                eestimator::BlockCrossValidation)
-
-  @assert !isa(problem, SimulationProblem) "not available for SimulationProblem"
   # problem info
+  @assert !isa(problem, SimulationProblem) "not available for SimulationProblem"
   probtype = typeof(problem)
   EP = probtype <: EstimationProblem ? true : false
+  
   sdata = EP ? data(problem) : sourcedata(problem)
   ovars = EP ? [v for (v,V) in variables(problem)] : outputvars(task(problem))
   loss  = eestimator.loss
--- a/src/errors/block_cross_validation.jl
+++ b/src/errors/block_cross_validation.jl
@@ -27,11 +27,16 @@ end
 BlockCrossValidation(sides; loss=Dict()) =
   BlockCrossValidation{typeof(sides)}(sides, loss)
 
-function error(solver::AbstractLearningSolver,
-               problem::LearningProblem,
+function error(solver::AbstractSolver,
+               problem::AbstractProblem,
                eestimator::BlockCrossValidation)
-  sdata = sourcedata(problem)
-  ovars = outputvars(task(problem))
+
+  @assert !isa(problem, SimulationProblem) "not available for SimulationProblem"
+  # problem info
+  probtype = typeof(problem)
+  EP = probtype <: EstimationProblem ? true : false
+  sdata = EP ? data(problem) : sourcedata(problem)
+  ovars = EP ? [v for (v,V) in variables(problem)] : outputvars(task(problem))
   loss  = eestimator.loss
   sides = eestimator.sides
   for var in ovars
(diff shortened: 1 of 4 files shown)


rmcaixeta:
@juliohm, about your initial points, I think it's more a difference about how the problems deal with CV. Because I ran a test with `CrossValidation()` and it's doing what other softwares do in kriging cross validation

Just to organize my thoughts, check if it makes sense: in LearningProblem there's all the workflow of dividing into training and tests sets, fit to features in the training set and predict something in the test set. In the EstimationProblems (OK, SK, UK, IDW...) we'll almost always have the hard data simultaneously as training and test set and an estimation domain with unknown information. All the geostats book I know just consider a k-fold CV in these cases using all the samples (usually a leave-one-out or some kind of leave-ball-out for close samples of same drillhole). It's a drawback, people question CV usefulness in these cases, but I don't know if there's some other way to deal with it. It's more an optional tool to help tuning parameters (search, variogram, etc) or select best methods, with no guarantee that these parameters are the best in the end.

In that sense, the code do that job and return the expected MSE after `error(solver, EstimationProblem, Validation(n))` for CrossValidation, BlockCrossValidation, LeaveBallOut and WeightedCrossValidation as they only deal with dataset/training set. I'll add specific Travis tests if you think that approach is okay

`DensityRatioValidation` seems to deal also with the estimation domain/test set, then adaptations would be necessary. But not sure if it'd be useful for EstimationProblem where we are estimating `:Z` using `:Z` (plus spatial coordinates), maybe it'd be better to let it available only for LearningProblems as it is right now. What do you think?

juliohm:
Thank you @rmcaixeta for the PR. I will try to review it over the weekend.

I don't know if I followed your comment about using all samples, but cross-validatory procedures require us to hide (or subset) samples in order to achieve an estimate of the error. I've submitted a paper last month to the Machine Learning journal on the CV methods implemented here, and I wish I could share it with you already. It should come out in the following months :pray: 

If other software out there use all the samples (the hard data) in every fold, may I ask how they compare the obtained results of the estimator with the ground truth? We can also try to setup a meeting to discuss this if it is too hard to explain in raw text on GitHub. Alternatively, if you are on the Zulip forum, we could try to write down some LaTeX there to make things more precise.

rmcaixeta:
Nice about the paper, I'll read it after published
About "all samples", I only mean by not dividing samples into training and test set, but considering everything as training set (where we know the ground truth). The rest remain the same (hiding one fold, and use the others to estimate the hidden one)

juliohm:
I think I get it. You are saying that other implementations will use all the hard data as training set as usual, but then they will try to estimate the entire domain containing all locations even when they are not part of the hard data locations. Is that it? So they post-process the results to compute losses for a subset of the domain as I understand it. Either way, I think we would need to modify the code in the PR to either (1) create sub-problems with sub-domains (preferred option for performance), or (2) evaluate losses only on hard data locations in the hidden fold (preferred for generality given that some solvers may not work with any kind of domain, and may require regular grid domains for example).

Please let me know if it makes sense. It would be awesome to support both EstimationProblem and LearningProblem in these methods.

rmcaixeta:
Isn't this already creating a sub-problem with subdomains?

```julia
probcall = getfield(Main, nameof(probtype))
thirdarg = EP ? Tuple(ovars) : task(problem
...
subproblem = probcall(train, hold, thirdarg)
```

juliohm:
The original `train` and `hold` variables before the PR were juts subsets of the `data` object, not  the domain object.

To instantiate a learning problem, we use two data objects: train data and test data.

In the case of CV:

- train data = remaining folds from original `data`
- test data = hidden fold from original `data`.

To instantiate an estimation problem, we use a data object and a domain object.

In the case of CV:

- data object = remaining folds from original `data`
- domain object = locations from original `domain` where the data from the hidden fold is found.

So although a data object has an underlying domain, and things may work if you pass it as the second argument to the EstimationProblem constructor, it is expecting a domain object. This is related to that discussion we had some time ago about instantiating estimation problems with two data objects. I think we should add an assertion in the problem constructors: https://github.com/JuliaEarth/GeoStats.jl/issues/131

Before we address this issue, we need to consider refactoring the various CV methods into a single implementation with the only difference being the method for creating folds: https://github.com/JuliaEarth/GeoStats.jl/issues/132

rmcaixeta:
Ah ok, got it
I was assuming everything before was ok
I'll close this for now then, let it rest for some time, the issue is getting big for the weekend

PR #70, opened 2020-12-16
Ellipsoid search and extra constraints

rmcaixeta:
PR plan:

- [x] `EllipsoidNeighborhood` object
- [x] `maxpercategory` and `maxperoctant` keyword options in some existing search method to validate them
- [x] generalize `maxpercategory` and `maxperoctant` filtering inside `BoundedSearch`
- [x] create additional tests
- [ ] review PR and rename BoundedSearch to FilteredSearch

drawing

### DIFF
--- a/src/distances.jl
+++ b/src/distances.jl
@@ -86,6 +86,15 @@ julia> aniso2distance([1.0,0.5,0.5], [π/2,0.0,0.0])
 function aniso2distance(semiaxes::AbstractVector, angles::AbstractVector;
   convention::Symbol=:TaitBryanExtr)
 
+  P, Λ = rotmat(semiaxes, angles, convention)
+  rule = rules[convention]
+  Q = P*Λ*P'
+  Mahalanobis(Q)
+end
+
+function rotmat(semiaxes::AbstractVector, angles::AbstractVector,
+  convention::Symbol=:TaitBryanExtr)
+
   N = length(semiaxes)
   @assert all(semiaxes .> 0) "semiaxes must be positive"
   @assert N ∈ [2,3] "dimension must be either 2 or 3"
@@ -110,9 +119,6 @@ function aniso2distance(semiaxes::AbstractVector, angles::AbstractVector;
 
   # rotation matrix
   P = angle_to_dcm(angles..., rule.order)[SOneTo(N),SOneTo(N)]
-
-  # ellipsoid matrix
-  Q = rule.extrinsic ? P*Λ*P' : P'*Λ*P
-
-  Mahalanobis(Q)
+  !rule.extrinsic && (P = P')
+  P, Λ
 end
--- a/src/distances.jl
+++ b/src/distances.jl
@@ -91,7 +91,7 @@ function aniso2distance(semiaxes::AbstractVector, angles::AbstractVector;
 end
 
 function rotmat(semiaxes::AbstractVector, angles::AbstractVector,
-                convention::Symbol=:TaitBryanExtr)
+                convention::Symbol=:TaitBryanExtr; rev=false)
   N = length(semiaxes)
   @assert all(semiaxes .> 0) "semiaxes must be positive"
   @assert N ∈ [2,3] "dimension must be either 2 or 3"
@@ -110,9 +110,12 @@ function rotmat(semiaxes::AbstractVector, angles::AbstractVector,
   !rule.radian && (angles = deg2rad.(angles))
   _0 = zero(eltype(angles))
   N == 2 && (angles = [angles[1], _0, _0])
-  intr = @. (rule.motion == :CW) & !rule.extrinsic
+  intr = @. (rule.motion == :CW)  & !rule.extrinsic
   extr = @. (rule.motion == :CCW) & rule.extrinsic
   angles[intr .| extr] *= -1
+
+  # for reverse extrinsic transformation
+  rule.extrinsic && rev && (angles *= -1)
 
   # rotation matrix
   P = angle_to_dcm(angles..., rule.order)[SOneTo(N),SOneTo(N)]
--- a/src/distances.jl
+++ b/src/distances.jl
@@ -84,17 +84,14 @@ julia> aniso2distance([1.0,0.5,0.5], [π/2,0.0,0.0])
 ```
 """
 function aniso2distance(semiaxes::AbstractVector, angles::AbstractVector;
-  convention::Symbol=:TaitBryanExtr)
-
+                        convention::Symbol=:TaitBryanExtr)
   P, Λ = rotmat(semiaxes, angles, convention)
-  rule = rules[convention]
   Q = P*Λ*P'
   Mahalanobis(Q)
 end
 
 function rotmat(semiaxes::AbstractVector, angles::AbstractVector,
-  convention::Symbol=:TaitBryanExtr)
-
+                convention::Symbol=:TaitBryanExtr)
   N = length(semiaxes)
   @assert all(semiaxes .> 0) "semiaxes must be positive"
   @assert N ∈ [2,3] "dimension must be either 2 or 3"
--- a/src/neighborhoods.jl
+++ b/src/neighborhoods.jl
@@ -10,6 +10,13 @@ A neighborhood of points in a spatial object.
 abstract type AbstractNeighborhood end
 
 """
+    AbstractBallNeighborhood
+
+A ball-like neighborhood of points in a spatial object.
+"""
+abstract type AbstractBallNeighborhood <: AbstractNeighborhood end
+
+"""
     isneighbor(neigh, xₒ, x)
 
 Tells whether or not the coordinates `x` are in the neighborhood
@@ -21,3 +28,4 @@ function isneighbor end
 # IMPLEMENTATIONS
 # ----------------
 include("neighborhoods/ball.jl")
+include("neighborhoods/ellipsoid.jl")
--- a/src/neighborhoods/ball.jl
+++ b/src/neighborhoods/ball.jl
@@ -7,7 +7,7 @@
 
 A ball neighborhood with `radius` and `metric`.
 """
-struct BallNeighborhood{T,M} <: AbstractNeighborhood
+struct BallNeighborhood{T,M} <: AbstractBallNeighborhood
   radius::T
   metric::M
 
@@ -36,20 +36,6 @@ metric(ball::BallNeighborhood) = ball.metric
 
 isneighbor(ball::BallNeighborhood, xₒ::AbstractVector, x::AbstractVector) =
   evaluate(ball.metric, xₒ, x) ≤ ball.radius
-
-"""
-    EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
-
-An ellipsoid neighborhood with `semiaxes` and `angles`. For 2D ellipses,
-there are two semiaxes and one rotation angle. For 3D ellipsoids, there are
-three semiaxes and three rotation angles. Different rotation conventions can be
-passed via the `convention` keyword argument. The list of conventions is
-available in the [ aniso2distance ](@ref) documentation.
-"""
-function EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
-  metric = aniso2distance(semiaxes, angles, convention=convention)
-  BallNeighborhood(one(eltype(semiaxes)), metric)
-end
 
 # ------------
 # IO methods
--- a/src/neighborhoods/ball.jl
+++ b/src/neighborhoods/ball.jl
@@ -38,14 +38,13 @@ isneighbor(ball::BallNeighborhood, xₒ::AbstractVector, x::AbstractVector) =
   evaluate(ball.metric, xₒ, x) ≤ ball.radius
 
 """
-    EllipsoidNeighborhood(semiaxes::AbstractVector, angles::AbstractVector;
-      convention::Symbol=:TaitBryanExtr)
+    EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
 
 An ellipsoid neighborhood with `semiaxes` and `angles`. For 2D ellipses,
 there are two semiaxes and one rotation angle. For 3D ellipsoids, there are
 three semiaxes and three rotation angles. Different rotation conventions can be
 passed via the `convention` keyword argument. The list of conventions is
-available in the `aniso2distance` documentation.
+available in the [ aniso2distance ](@ref) documentation.
 """
 function EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
   metric = aniso2distance(semiaxes, angles, convention=convention)
--- a/src/neighborhoods/ball.jl
+++ b/src/neighborhoods/ball.jl
@@ -37,6 +37,16 @@ metric(ball::BallNeighborhood) = ball.metric
 isneighbor(ball::BallNeighborhood, xₒ::AbstractVector, x::AbstractVector) =
   evaluate(ball.metric, xₒ, x) ≤ ball.radius
 
+"""
+    EllipsoidNeighborhood(semiaxes::AbstractVector, angles::AbstractVector;
+      convention::Symbol=:TaitBryanExtr)
+
+An ellipsoid neighborhood with `semiaxes` and `angles`. For 2D ellipses,
+there are two semiaxes and one rotation angle. For 3D ellipsoids, there are
+three semiaxes and three rotation angles. Different rotation conventions can be
+passed via the `convention` keyword argument. The list of conventions is
+available in the `aniso2distance` documentation.
+"""
 function EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
   metric = aniso2distance(semiaxes, angles, convention=convention)
   BallNeighborhood(1.0, metric)
--- a/src/neighborhoods/ball.jl
+++ b/src/neighborhoods/ball.jl
@@ -48,7 +48,7 @@ available in the [ aniso2distance ](@ref) documentation.
 """
 function EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
   metric = aniso2distance(semiaxes, angles, convention=convention)
-  BallNeighborhood(1.0, metric)
+  BallNeighborhood(one(eltype(semiaxes)), metric)
 end
 
 # ------------
--- a/src/neighborhoods/ball.jl
+++ b/src/neighborhoods/ball.jl
@@ -37,6 +37,11 @@ metric(ball::BallNeighborhood) = ball.metric
 isneighbor(ball::BallNeighborhood, xₒ::AbstractVector, x::AbstractVector) =
   evaluate(ball.metric, xₒ, x) ≤ ball.radius
 
+function EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
+  metric = aniso2distance(semiaxes, angles, convention=convention)
+  BallNeighborhood(1.0, metric)
+end
+
 # ------------
 # IO methods
 # ------------
--- a/src/neighborhoods/ellipsoid.jl
+++ b/src/neighborhoods/ellipsoid.jl
@@ -1,0 +1,45 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
+
+An ellipsoid neighborhood with `semiaxes` and `angles`. For 2D ellipses,
+there are two semiaxes and one rotation angle. For 3D ellipsoids, there are
+three semiaxes and three rotation angles. Different rotation conventions can be
+passed via the `convention` keyword argument. The list of conventions is
+available in the [ aniso2distance ](@ref) documentation.
+"""
+struct EllipsoidNeighborhood <: AbstractBallNeighborhood
+  semiaxes::AbstractVector
+  angles::AbstractVector
+  convention::Symbol
+  metric::Mahalanobis
+
+  function EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
+    metric = aniso2distance(semiaxes, angles, convention=convention)
+    new(semiaxes, angles, convention, metric)
+  end
+end
+
+radius(ellips::EllipsoidNeighborhood) = one(eltype(ellips.semiaxes))
+
+metric(ellips::EllipsoidNeighborhood) = ellips.metric
+
+isneighbor(ellips::EllipsoidNeighborhood, xₒ::AbstractVector, x::AbstractVector) =
+  evaluate(ellips.metric, xₒ, x) ≤ radius(ellips)
+
+# ------------
+# IO methods
+# ------------
+function Base.show(io::IO, ellips::EllipsoidNeighborhood)
+  print(io, "EllipsoidNeighborhood")
+end
+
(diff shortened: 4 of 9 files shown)


juliohm:
Awesome. Will review it soon...

src/neighborhoods/ball.jl
```diff
@@ -37,6 +37,11 @@ metric(ball::BallNeighborhood) = ball.metric
 isneighbor(ball::BallNeighborhood, xₒ::AbstractVector, x::AbstractVector) =
   evaluate(ball.metric, xₒ, x) ≤ ball.radius
 
+function EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
+  metric = aniso2distance(semiaxes, angles, convention=convention)
+  BallNeighborhood(1.0, metric)
+end
+
```
juliohm, line 44: What do you think of moving this to a separate source file `ellipsoid.jl` in the same folder? I know it is just a helper function, but maybe we could place it in a separate file and write a docstring for it pointing to the `aniso2distance` docstring.

juliohm (changes requested):
That is nice. We need tests and a docstring for the `EllipsoidNeighborhood`.

I was also wondering if we would benefit from making this a completely separate type instead of a helper function that just returns a `BallNeighborhood`. Can you think of situations where we would like to behave differently from a `BallNeighborhood`?

src/neighborhoods/ball.jl
juliohm: So the question of moving to a separate file can only be addressed after we brainstorm possible use cases where it is important to differentiate between `EllipsoidNeighborhood` and `BallNeighborhood`. If it is important to different and dispatch different methods, we need a separate type `EllipsoidNeighborhood` in a separate source file. It if it is not important, I can leave the definition here as a helper function that returns a `BallNeighborhood`.

rmcaixeta: I was initially inclined to think that it needed to be a separate structure. But now I think it's really just a `BallNeighborhood` with different metric, where the helper function is useful to simplify the call. The restrictions can also be applied to both, including octants. I'm just not totally sure if I can make the octants restriction using the `Mahalanobis.qmat` as it is (I need to transform the coordinates in order that the semiaxes match xyz axis).

rmcaixeta:
For the sequence, I thought about a `filterneigh` functions that is called in the end of `search(args...)` if some restriction is called. Something like that:

```julia
function search(xₒ, method::NeighborhoodSearch{O,N,T}) where {O,N<:BallNeighborhood,T}
	# ... do everything as already coded to get neighbor indices ...
	method.restrictions && filterneigh!(indices, method)
	indices
end
```

src/neighborhoods/ball.jl
juliohm: Yes, I also cannot see a case where we would need to differentiate the behavior of `EllipsoidNeighborhood` from the behavior of `BallNeighborhood`, so I think we are fine with the helper function like this. We just need a docstring and tests 👍🏽 

If you can describe the octant restriction that would help understand your concerns with coordinates transformations.

juliohm:
Oh so you are thinking more generally to add these options to all kinds of search methods? Maybe we could create a meta-search struct like the `BoundedSearch` one that take a search method and post-processes the result with additional parameters as suggested.

src/neighborhoods/ball.jl
```diff
@@ -37,6 +37,21 @@ metric(ball::BallNeighborhood) = ball.metric
...
 
+"""
+    EllipsoidNeighborhood(semiaxes::AbstractVector, angles::AbstractVector;
+      convention::Symbol=:TaitBryanExtr)
+
+An ellipsoid neighborhood with `semiaxes` and `angles`. For 2D ellipses,
+there are two semiaxes and one rotation angle. For 3D ellipsoids, there are
+three semiaxes and three rotation angles. Different rotation conventions can be
+passed via the `convention` keyword argument. The list of conventions is
+available in the `aniso2distance` documentation.
```
juliohm, line 48: Awesome docstring. You can use `[`aniso2distance`](@ref)` to make it a link in the docs.

juliohm: GitHub is messing up with the single quotes. Check other example of `@ref` in the codebase for links in docstrings 👌🏽

rmcaixeta: Nice trick, I didn't know that

src/neighborhoods/ball.jl
```diff
@@ -37,6 +37,21 @@ metric(ball::BallNeighborhood) = ball.metric
 isneighbor(ball::BallNeighborhood, xₒ::AbstractVector, x::AbstractVector) =
   evaluate(ball.metric, xₒ, x) ≤ ball.radius
 
+"""
+    EllipsoidNeighborhood(semiaxes::AbstractVector, angles::AbstractVector;
+      convention::Symbol=:TaitBryanExtr)
```
juliohm, line 42: I would eliminate the type annotations in the docstring:

```julia
EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
```

rmcaixeta:
> Oh so you are thinking more generally to add these options to all kinds of search methods? Maybe we could create a meta-search struct like the `BoundedSearch` one that takes a search method and post-processes the result with additional parameters as suggested.

It might be useful for other searches too. I'll do a simple function and tests here first to see if it works. Then we move/adapt it to make it more generic if necessary, sounds good?

juliohm:
Sounds awesome. We can iterate on the design, it is a quite important building block to get right 👍🏽

rmcaixeta:
Enough for today. I'll test tomorrow to check if the code is really doing what I want. But already gives an idea of what's necessary

src/neighborsearch/neighborhood.jl
```diff
@@ -63,8 +76,69 @@ function search(xₒ::AbstractVector, method::NeighborhoodSearch{O,N,T};
...
+  # get distances and give priority to closest neighbors according to ordermetric
+  dists   = colwise(ometric, xₒ, coordinates(object,inds))
+  sortids = sortperm(dists)
+  inds    = inds[sortids]
+
+  # key and octant constraints
+  keys   = usekey ? unique(view(values(object), inds, maxkey[1])) : nothing
+  dkeys  = usekey ? Dict(zip(keys, 1:length(keys)))               : nothing
+  ctkeys = usekey ? zeros(Int, length(keys))                      : nothing
+  octs   = useoct ? zeros(Int, 8)                                 : nothing
```
juliohm, line 110: We need to refactor this code to make it more easy to follow. I see that there is some logic to initialize some auxiliary variables in the filtering that comes later. Am I correct that there two disjoint filtering operations happening in the loop? If that is the case, I would create two separate sections in the function body, one for the filtering with `usekey` and another section for the filtering with `useoct`. The second section could use the result of the first section for performance reasons and clarity.

src/neighborsearch/neighborhood.jl
```diff
@@ -63,8 +76,69 @@ function search(xₒ::AbstractVector, method::NeighborhoodSearch{O,N,T};
...
+
+  neighbors = Vector{Int}()
+  ctneighs  = 0
+
+  @inbounds for ind in inds
+    key = usekey ? values(object)[ind,maxkey[1]]                    : nothing
+    oct = useoct ? getoct(rotmat * (coordinates(object,ind) .- xₒ)) : nothing
+
+    usekey && ctkeys[dkeys[key]] >= maxkey[2] && continue
+    useoct && octs[oct] >= maxoct             && continue
```
juliohm, line 120: So this is where the filtering is happening, right? It is very hard to understand what exactly is going on, and so I suggest to add some comments to explain what your intentions are with the code written. I like the formatting btw, very nice to see a PR with with everything aligned properly 💯

src/neighborsearch/neighborhood.jl
```diff
@@ -63,8 +76,69 @@ function search(xₒ::AbstractVector, method::NeighborhoodSearch{O,N,T};
...
+
+# get octant code of transformed coordinates
+function getoct(coords::AbstractVector)
+  N = size(coords,1)
+  i = (coords .< 0) .+ 1
+
+  octid = N == 2 ? reshape(1:4,(2,2)) : reshape(1:8,(2,2,2))
+  oct   = N == 2 ? octid[i[1],i[2]]   : octid[i[1],i[2],i[3]]
+  oct
+end
```
juliohm, line 144: Sorry my lack of understanding here, but would you mind explaining the filtering operations you have in mind in a high-level language, possibly with illustrations? I could provide better feedback that way.

juliohm (changes requested):
Hi @rmcaixeta , thanks for working on such nice features. I am looking forward to a version of the PR with the details sorted out. After looking carefully at the code, I think we need a separate meta-search method a la `BoundedSearch` that takes any of the existing search methods and provides this additional step you suggested with octant and other filtering operations. Perhaps we could even rename `BoundedSearch` to `FilteredSearch` and generalize it with more options. What do you think?

So basically, after you incorporate the specific comments I left here, we should decide if the functionality goes to a separate and new `FilteredSearch` or if we rename `BoundedSearch` to `FilteredSearch` and add the extra options.

src/neighborsearch/neighborhood.jl
rmcaixeta: Here the sketch of the idea

text1203-8

src/neighborsearch/neighborhood.jl
```diff
@@ -11,12 +11,18 @@ struct NeighborhoodSearch{O,N,T} <: NeighborSearchMethod
   # input fields
   object::O
   neigh::N
+  maxneighs::Int
+  maxperoct::Int
+  maxperkey::Tuple{Symbol,Int}
```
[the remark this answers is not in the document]
rmcaixeta: Using a dictionary as we talked earlier, to include more categories if desired. Maybe could be a vector of pairs too

rmcaixeta:
Made some checks here, I need some extra information to make octants really work
For octants, I need to transform coordinates to make cartesian axes match ellips. semiaxes
I was trying to run it with `aniso2distance` matrix `P*Λ*P'`. But it doesn't work, I need to have only `P'`

I'm thinking of making `EllipsoidNeighborhood` a separate struct instead of just a helper function. And store `P'` as a parameter. Seems ok to you?

juliohm:
Yes, I think it makes sense to have a separate `EllipsoidNeighborhood` to save the input parameters `semiaxes` and `angles`. That way any algorithm that uses a neighborhood will be able to recover this information.

Another thing we may need to do at some point in the future is to refactor the `aniso2distance` to expose the rotation operations, etc. I think these could live in Meshes.jl, so we don't need to bother right now.

src/distances.jl
```diff
@@ -104,15 +110,15 @@ function aniso2distance(semiaxes::AbstractVector, angles::AbstractVector;
   !rule.radian && (angles = deg2rad.(angles))
   _0 = zero(eltype(angles))
   N == 2 && (angles = [angles[1], _0, _0])
-  intr = @. (rule.motion == :CW) & !rule.extrinsic
+  intr = @. (rule.motion == :CW)  & !rule.extrinsic
```
juliohm, line 113: Extra space added by accident?

src/distances.jl
```diff
@@ -104,15 +110,15 @@ function aniso2distance(semiaxes::AbstractVector, angles::AbstractVector;
...
   extr = @. (rule.motion == :CCW) & rule.extrinsic
   angles[intr .| extr] *= -1
 
+  # for reverse extrinsic transformation
+  rule.extrinsic && rev && (angles *= -1)
+
```
juliohm, line 119: So you are replacing the transposition by a flip of sign in the angle? Can you comment on this change?

src/neighborhoods.jl
```diff
@@ -9,6 +9,13 @@ A neighborhood of points in a spatial object.
 """
 abstract type AbstractNeighborhood end
 
+"""
+    AbstractBallNeighborhood
+
+A ball-like neighborhood of points in a spatial object.
+"""
+abstract type AbstractBallNeighborhood <: AbstractNeighborhood end
```
juliohm, line 17: Can we just keep the original `AbstractNeighborhood` type? Why we need an extra level of abstraction? Are you doing because of dispatch issues somewhere else?

src/neighborhoods/ellipsoid.jl
```diff
@@ -0,0 +1,45 @@
...
+    new(semiaxes, angles, convention, metric)
+  end
+end
+
+radius(ellips::EllipsoidNeighborhood) = one(eltype(ellips.semiaxes))
+
+metric(ellips::EllipsoidNeighborhood) = ellips.metric
+
+isneighbor(ellips::EllipsoidNeighborhood, xₒ::AbstractVector, x::AbstractVector) =
+  evaluate(ellips.metric, xₒ, x) ≤ radius(ellips)
```
juliohm, line 31: Awesome implementation 💯

src/distances.jl
rmcaixeta: For octants, I need to reverse rotation in order the semiaxes match cartesian axes. Intrinsic rotations can be reversed only by the transpose of rotation matrix. Extrinsic case need to invert sign of the angles, that's why I included this rev kwarg and operation

src/neighborhoods.jl
rmcaixeta: I added the AbstractBallNeighborhood because there is a NeighborhoodSearch and a search method specific for Union(BallNeighborhood, EllipsoidNeighborhood)

src/distances.jl
rmcaixeta: Thanks. I'm without git in the machine I'm using this week. On monday I'll commit the adjusts

src/neighborhoods.jl
juliohm: I don't know if I understood the comment. I thought this PR is adding the `EllipsoidNeighborhood` right? How there is an existing method with it? If you can clarify why we need an extra abstract type, that would be helpful 👍🏽

rmcaixeta: "method" might not be the correct term. Putting in other words, the `EllipsoidNeighborhood` is a `BallNeighborhood` with Mahalanobis metric as we discussed initially. The `AbstractBallNeighborhood` was defined to also consider Ellipsoid into two Ball specific calls inside `src/neighborsearch/neighborhood.jl`. Check the changes in that file, it might get clearer why I made it

juliohm: I will look into it more carefully later. I think I got your point. You are making a common parent type so that the KD tree search is triggered for both BallNeighborhood and EllipoidNeighborhood right? Since these two are types of "Balls". I wonder if we should treat them completely separate cases now that the EllipsoidNeighborhood is a separate struct. Perhaps avoiding the common parent AbstractBallNeighborhood and just using AbstractNeighborhood? That would just need a new entry in  the `Union{}`of the appropriate method I guess. I will think more about it too.

src/neighborsearch/bounded.jl
```diff
@@ -3,39 +3,122 @@
...
 struct BoundedSearch{M<:NeighborSearchMethod} <: BoundedNeighborSearchMethod
   method::M
-  nmax::Int
+  maxneighbors::Int
+  maxpercategory::Dict{Symbol,Int}
+  maxperoctant::Int
+  ordermetric::Metric
+
+  function BoundedSearch{M}(method::M, maxneighbors=0; maxpercategory=Dict(),
+                            maxperoctant=0, ordermetric=Euclidean()) where {M}
```
juliohm, line 21: We don't usually add default values in inner constructors, nor specify the types of the arguments. At least that is the patterns I've seen most often out there. Could you please rewrite this as:

```julia
function BoundedSearch{M}(method, maxneighbors, maxpercategory, maxperoctant, ordermetric) where {M}
```

and then move the logic about default values and keywords to outer constructors? This is the pattern we are following the rest of the codebase.

src/neighborsearch/bounded.jl
```diff
@@ -3,39 +3,122 @@
...
+  # get reference objects and initial neighbors
+  meth = method.method
+  obj  = object(method)
+  N    = ncoords(obj)
+  inds = search(xₒ, meth, mask=mask)
+
+  # get distances and give priority to closest neighbors according to ordermetric
+  ometric = method.ordermetric
+  dists   = colwise(ometric, xₒ, coordinates(obj,inds))
+  sortids = sortperm(dists)
```
juliohm, line 44: Rename this variable to `sorted` to not confuse with `sortinds`

src/neighborsearch/bounded.jl
```diff
@@ -3,39 +3,122 @@
...
+  inds    = inds[sortids]
+
+  # read constraints
+  maxk   = method.maxneighbors
+  maxoct = method.maxperoctant
+  maxcat = method.maxpercategory
```
juliohm, line 50: I would move all these lines where we are simply accessing fields of the method to the top of the implementation. After all fields are read, we can then start the logic in a separate section, step by step. Appreciate if you can refactor this.

src/neighborsearch/bounded.jl
```diff
@@ -3,39 +3,122 @@
...
+  usek   = maxk   > 0
+  useoct = maxoct > 0
+  usecat = length(maxcat) > 0
```
juliohm, line 53: If these variables are only used once in the conditional statements, maybe they shouldn't exist. For example, in the next section below we could just write

```julia
# handle octant constraint
if maxoct > 0`
  ...
end
```

src/neighborsearch/bounded.jl
```diff
@@ -3,39 +3,122 @@
...
+
+  # initialize octant restriction
```
juliohm, line 55: let's use the term `constraint` in place of `restriction`. In math the name `restriction` is used for something else: when you restrict a function to a subdomain.

src/neighborsearch/bounded.jl
juliohm: Oh I see that some of these are used more than once. If you feel that it would be nicer to organize them in a section like this, that is fine too. Just make sure that the section is highlighted and not mixed with other parts of the function body.

src/neighborsearch/bounded.jl
```diff
@@ -3,39 +3,122 @@
...
+    nneigh += 1
+    if usek
+      neighbors[nneigh] = ind
+    else
+      push!(neighbors, ind)
+    end
+    useoct && (octs[oct] += 1)
+    if usecat
+      for col in keys(maxcat)
+        ctcatgs[col][cat[col]] += 1
```
juliohm, line 101: This line is an indication that we need to refactor this code somehow. The indexing pattern is too complex, so we probably need more auxiliary variables somewhere to make the code more readable.

src/neighborsearch/bounded.jl
```diff
@@ -3,39 +3,122 @@
...
   end
 
   nneigh
 end
+
+# get octant code of centered coordinates
+function getoct(coords::AbstractVector)
+  # get dimensions
+  N    = size(coords,1)
+  dims = Tuple([2 for x in 1:N])
```
juliohm, line 116: The efficient method to construct this tuple is `ntuple(i->2, N)`

src/neighborsearch/bounded.jl
```diff
@@ -3,39 +3,122 @@
...
+    usek && nneigh >= maxk && break
   end
 
   nneigh
 end
+
+# get octant code of centered coordinates
+function getoct(coords::AbstractVector)
+  # get dimensions
+  N    = size(coords,1)
```
juliohm, line 115: missing space after comma

src/neighborsearch/bounded.jl
```diff
@@ -3,39 +3,122 @@
...
     end
+
+    # if maxneigh reached, stop
+    usek && nneigh >= maxk && break
   end
 
   nneigh
 end
+
+# get octant code of centered coordinates
```
juliohm, line 112: Can you clarify what this function does? Imagine that someone is reading this code for the first time, and that this person has intermediate programming skills. They should still be able to see the geometric algorithm you have in mind. We need to document this functionality better so that people can maintain and fix bugs in the future in case there are any.

src/neighborsearch/neighborhood.jl
```diff
@@ -17,7 +17,11 @@ struct NeighborhoodSearch{O,N,T} <: NeighborSearchMethod
 end
 
 function NeighborhoodSearch(object::O, neigh::N) where {O,N}
-  tree = if neigh isa BallNeighborhood
+  if neigh isa EllipsoidNeighborhood
+    N1, N2 = ncoords(object), size(metric(neigh).qmat,1)
+    @assert  N1 == N2  "data and ellipse/ellipsoid must have the same dimensions"
+ end
```
juliohm, line 23: Indentation issue

test/neighborhoods.jl
```diff
@@ -26,4 +26,26 @@
...
+    n = EllipsoidNeighborhood([2.0, 0.75, 0.25], [45, -45, 0], convention=:GSLIB)
+
+    @test isneighbor(n, [0.,0.,0.], 1.9 .* [0.5 ;   0.5; -√2/2])
+    @test isneighbor(n, [0.,0.,0.], 0.7 .* [√2/2; -√2/2;   0.0])
+    @test isneighbor(n, [0.,0.,0.], 0.2 .* [0.5 ;   0.5;  √2/2])
+
+    @test !isneighbor(n, [0.,0.,0.], 2.1 .* [0.5 ;   0.5; -√2/2])
+    @test !isneighbor(n, [0.,0.,0.], 0.8 .* [√2/2; -√2/2;   0.0])
+    @test !isneighbor(n, [0.,0.,0.], 0.3 .* [0.5 ;   0.5;  √2/2])
+  end
```
juliohm, line 50: Awesome section of tests. I think we just need some comments to help the reader understand what is the configuration you have in mind in each of these tests.

test/neighborsearch.jl
```diff
@@ -103,5 +104,59 @@
...
+
+		# 3-D case
+		dims = (10, 10, 10)
+		𝒟 = georef((K1=rand(1:2, dims), K2=rand(1:2, dims)))
+		S1 = NeighborhoodSearch(𝒟, EllipsoidNeighborhood([10. ,8. ,5.],[45, 45, 0]))
+		S2 = KNearestSearch(𝒟, 64)
+
+		args   = iprod([S1, S2], maxk)
+		kwargs = [(maxperoctant=x[1], maxpercategory=x[2]) for x in iprod(maxoct, maxcat)]
+		boundedtests(args, kwargs)
```
juliohm, line 160: Indentation issue.

juliohm (changes requested):
Another round of review, I think I am starting to see the algorithm you have in mind in the code itself. More comments and illustrations can help a lot. We had a chat in Zulip, and I am still finding it difficult to grasp the intention in the code. Imagine someone who hasn't had the chance to chat with you, they should be able to read the code and say: Aha! Makes sense! Clever!

juliohm:
@rmcaixeta did you have a chance to finish up this PR? It would be nice to add the ellipsoid search.

rmcaixeta:
Hey @juliohm . Got no time yet. Hope to get back to this PR on the next weekend or some days after

juliohm:
No worries. When you come back to it, could you please split this PR into two separate PRs, one for the new Ellisoid search method, and another for the additional set of constraints? The former is non breaking and we can quickly merge and release a new version of the project. The latter requires fixes downstream in solvers that use `BoundedSearch`.

juliohm:
@rmcaixeta what do you think of this one? Should we start a fresh PR with the extra constraints meta searcher or adapt this one?

rmcaixeta:
I think it will be easier to start a new one.  I'll update my local files with your reviews later and then I start a new one. There are some other details I want to improve on that later

juliohm:
@rmcaixeta I think we will need to close this one given the enormous amount of change in the codebase lately. This PR should go to Meshes.jl now. 👍🏽

PR #75, opened 2021-02-08
Add EllipsoidNeighborhood

rmcaixeta:
Splitting #70 PR to deal only with `EllipsoidNeighborhood` here

### DIFF
--- a/src/neighborhoods/ellipsoid.jl
+++ b/src/neighborhoods/ellipsoid.jl
@@ -1,0 +1,45 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
+
+An ellipsoid neighborhood with `semiaxes` and `angles`. For 2D ellipses,
+there are two semiaxes and one rotation angle. For 3D ellipsoids, there are
+three semiaxes and three rotation angles. Different rotation conventions can be
+passed via the `convention` keyword argument. The list of conventions is
+available in the [ aniso2distance ](@ref) documentation.
+"""
+struct EllipsoidNeighborhood <: AbstractBallNeighborhood
+  semiaxes::AbstractVector
+  angles::AbstractVector
+  convention::Symbol
+  metric::Mahalanobis
+
+  function EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
+    metric = aniso2distance(semiaxes, angles, convention=convention)
+    new(semiaxes, angles, convention, metric)
+  end
+end
+
+radius(ellips::EllipsoidNeighborhood) = one(eltype(ellips.semiaxes))
+
+metric(ellips::EllipsoidNeighborhood) = ellips.metric
+
+isneighbor(ellips::EllipsoidNeighborhood, xₒ::AbstractVector, x::AbstractVector) =
+  evaluate(ellips.metric, xₒ, x) ≤ radius(ellips)
+
+# ------------
+# IO methods
+# ------------
+function Base.show(io::IO, ellips::EllipsoidNeighborhood)
+  print(io, "EllipsoidNeighborhood")
+end
+
+function Base.show(io::IO, ::MIME"text/plain", ellips::EllipsoidNeighborhood)
+  println(io, "EllipsoidNeighborhood")
+  println(io, "  semiaxes: ", ellips.semiaxes)
+  println(io, "  angles: ", ellips.angles)
+  print(  io, "  rotation convention: ", ellips.convention)
+end
--- a/src/neighborhoods/ellipsoid.jl
+++ b/src/neighborhoods/ellipsoid.jl
@@ -9,7 +9,7 @@ An ellipsoid neighborhood with `semiaxes` and `angles`. For 2D ellipses,
 there are two semiaxes and one rotation angle. For 3D ellipsoids, there are
 three semiaxes and three rotation angles. Different rotation conventions can be
 passed via the `convention` keyword argument. The list of conventions is
-available in the [ aniso2distance ](@ref) documentation.
+available in the [aniso2distance](@ref) documentation.
 """
 struct EllipsoidNeighborhood <: AbstractBallNeighborhood
   semiaxes::AbstractVector
--- a/src/neighborhoods/ellipsoid.jl
+++ b/src/neighborhoods/ellipsoid.jl
@@ -17,14 +17,14 @@ struct EllipsoidNeighborhood{S,A,C,M} <: AbstractBallNeighborhood
   convention::C
   metric::M
 
-  function EllipsoidNeighborhood{S,A,C,M}(semiaxes, angles; convention) where {S,A,C,M}
+  function EllipsoidNeighborhood{S,A,C,M}(semiaxes, angles, convention) where {S,A,C,M}
     metric = aniso2distance(semiaxes, angles, convention=convention)
     new(semiaxes, angles, convention, metric)
   end
 end
 
 EllipsoidNeighborhood(semiaxes::S, angles::A; convention::C=:TaitBryanExtr) where {S,A,C} =
-  EllipsoidNeighborhood{S,A,C,Mahalanobis}(semiaxes, angles; convention)
+  EllipsoidNeighborhood{S,A,C,Mahalanobis}(semiaxes, angles, convention)
 
 metric(ellips::EllipsoidNeighborhood) = ellips.metric
 
--- a/src/neighborhoods/ellipsoid.jl
+++ b/src/neighborhoods/ellipsoid.jl
@@ -11,24 +11,25 @@ three semiaxes and three rotation angles. Different rotation conventions can be
 passed via the `convention` keyword argument. The list of conventions is
 available in the [aniso2distance](@ref) documentation.
 """
-struct EllipsoidNeighborhood <: AbstractBallNeighborhood
-  semiaxes::AbstractVector
-  angles::AbstractVector
-  convention::Symbol
-  metric::Mahalanobis
+struct EllipsoidNeighborhood{S,A,C,M} <: AbstractBallNeighborhood
+  semiaxes::S
+  angles::A
+  convention::C
+  metric::M
 
-  function EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
+  function EllipsoidNeighborhood{S,A,C,M}(semiaxes, angles; convention) where {S,A,C,M}
     metric = aniso2distance(semiaxes, angles, convention=convention)
     new(semiaxes, angles, convention, metric)
   end
 end
 
-radius(ellips::EllipsoidNeighborhood) = one(eltype(ellips.semiaxes))
+EllipsoidNeighborhood(semiaxes::S, angles::A; convention::C=:TaitBryanExtr) where {S,A,C} =
+  EllipsoidNeighborhood{S,A,C,Mahalanobis}(semiaxes, angles; convention)
 
 metric(ellips::EllipsoidNeighborhood) = ellips.metric
 
 isneighbor(ellips::EllipsoidNeighborhood, xₒ::AbstractVector, x::AbstractVector) =
-  evaluate(ellips.metric, xₒ, x) ≤ radius(ellips)
+  evaluate(ellips.metric, xₒ, x) ≤ one(eltype(ellips.semiaxes))
 
 # ------------
 # IO methods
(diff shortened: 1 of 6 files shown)


src/neighborsearch/neighborhood.jl
```diff
@@ -17,7 +17,11 @@ struct NeighborhoodSearch{O,N,T} <: NeighborSearchMethod
 end
 
 function NeighborhoodSearch(object::O, neigh::N) where {O,N}
-  tree = if neigh isa BallNeighborhood
+  if neigh isa EllipsoidNeighborhood
+    N1, N2 = ncoords(object), size(metric(neigh).qmat,1)
+    @assert  N1 == N2  "data and ellipse/ellipsoid must have the same dimensions"
+ end
```
juliohm, line 23: Do we really need this assertion here? If we want to add it, I think we need to first provide  public api to access the dimension of the ellipsoid neighborhood. The `metric(neigh).qmat` part is an internal detail.

juliohm (changes requested):
Amazing PRs as usual 💯 I wonder if we really need that assertion code there? Perhaps we could add in a future PR? We don't have a public API for the ellipsoid neighborhood dimension yet.

I like your design to create a separate struct for the Ellipsoid case. It is a ball neighborhood too but it is so specialized that it deserves a struct for dispatch. The `AbstractBallNeighborhood` also seems reasonable.

rmcaixeta:
Hey @juliohm 
Thanks for the quick review. I removed the assertion on the last commit
Then all the other issues of the last PR is about `BoundedSearch`, I'll work on them later
Are the tests crashing because of the visual tests?

src/neighborhoods/ellipsoid.jl
```diff
@@ -0,0 +1,45 @@
...
+# ------------------------------------------------------------------
+
+"""
+    EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
+
+An ellipsoid neighborhood with `semiaxes` and `angles`. For 2D ellipses,
+there are two semiaxes and one rotation angle. For 3D ellipsoids, there are
+three semiaxes and three rotation angles. Different rotation conventions can be
+passed via the `convention` keyword argument. The list of conventions is
+available in the [ aniso2distance ](@ref) documentation.
```
juliohm, line 12:
```suggestion
available in the [aniso2distance](@ref) documentation.
```

src/neighborhoods/ellipsoid.jl
```diff
@@ -0,0 +1,45 @@
...
+  convention::Symbol
+  metric::Mahalanobis
+
+  function EllipsoidNeighborhood(semiaxes, angles; convention=:TaitBryanExtr)
+    metric = aniso2distance(semiaxes, angles, convention=convention)
+    new(semiaxes, angles, convention, metric)
+  end
+end
+
+radius(ellips::EllipsoidNeighborhood) = one(eltype(ellips.semiaxes))
```
juliohm, line 26: The radius function doesn't make much sense for the Ellipsoid. Can we remove this here and modify the search method slightly to handle this instead?

src/neighborhoods/ellipsoid.jl
```diff
@@ -0,0 +1,45 @@
...
+there are two semiaxes and one rotation angle. For 3D ellipsoids, there are
+three semiaxes and three rotation angles. Different rotation conventions can be
+passed via the `convention` keyword argument. The list of conventions is
+available in the [aniso2distance](@ref) documentation.
+"""
+struct EllipsoidNeighborhood <: AbstractBallNeighborhood
+  semiaxes::AbstractVector
+  angles::AbstractVector
+  convention::Symbol
+  metric::Mahalanobis
```
juliohm, line 18: I think we need to add type parameters here to avoid abstract types. For example:

```julia
struct EllipsoidNeighborhood{S,A,C,M} <: AbstractBallNeighborhood
  semiaxes::S
  angles::A
  convention::C
  metric::M
end
```

src/neighborhoods/ellipsoid.jl
```diff
@@ -0,0 +1,45 @@
...
+# ------------
+function Base.show(io::IO, ellips::EllipsoidNeighborhood)
+  print(io, "EllipsoidNeighborhood")
+end
+
+function Base.show(io::IO, ::MIME"text/plain", ellips::EllipsoidNeighborhood)
+  println(io, "EllipsoidNeighborhood")
+  println(io, "  semiaxes: ", ellips.semiaxes)
+  println(io, "  angles: ", ellips.angles)
+  print(  io, "  rotation convention: ", ellips.convention)
```
juliohm, line 44:
```suggestion
  print(  io, "  convention: ", ellips.convention)
```

juliohm (changes requested):
Attached is a second round of reviews. I think the errors in the test suite are unrelated. Unfortunately the MLJ.jl stack and the Plots.jl stack are constantly changing and breaking. We can fix that later after this PR is ready and merged.

rmcaixeta:
> Attached is a second round of reviews. I think the errors in the test suite are unrelated. Unfortunately the MLJ.jl stack and the Plots.jl stack are constantly changing and breaking. We can fix that later after this PR is ready and merged.

Thanks. Changes made (not sure if parametric types initialization is 100% ok, but it's running well here)

src/neighborhoods/ellipsoid.jl
```diff
@@ -0,0 +1,46 @@
...
+  metric::M
+
+  function EllipsoidNeighborhood{S,A,C,M}(semiaxes, angles; convention) where {S,A,C,M}
+    metric = aniso2distance(semiaxes, angles, convention=convention)
+    new(semiaxes, angles, convention, metric)
+  end
+end
+
+EllipsoidNeighborhood(semiaxes::S, angles::A; convention::C=:TaitBryanExtr) where {S,A,C} =
+  EllipsoidNeighborhood{S,A,C,Mahalanobis}(semiaxes, angles; convention)
```
juliohm, line 27:
```suggestion
  function EllipsoidNeighborhood{S,A,C,M}(semiaxes, angles convention) where {S,A,C,M}
    metric = aniso2distance(semiaxes, angles, convention=convention)
    new(semiaxes, angles, convention, metric)
  end
end

EllipsoidNeighborhood(semiaxes::S, angles::A; convention::C=:TaitBryanExtr) where {S,A,C} =
  EllipsoidNeighborhood{S,A,C,Mahalanobis}(semiaxes, angles, convention)
```

src/neighborsearch/neighborhood.jl
```diff
@@ -54,8 +55,11 @@ end
 
 # search method for ball neighborhood
 function search(xₒ::AbstractVector, method::NeighborhoodSearch{O,N,T};
-                mask=nothing) where {O,N<:BallNeighborhood,T}
-  inds = inrange(method.tree, xₒ, radius(method.neigh))
+                mask=nothing) where {O,N<:AbstractBallNeighborhood,T}
+  neigh = method.neigh
+  range = N <: BallNeighborhood ? radius(neigh) : one(eltype(neigh.semiaxes))
```
juliohm, line 60: We need provide a public API to recover this "one" for the ellipsoid case. Or perhaps we need to modify the original BallNeighborhood so that the "one" can be used too without the need to querying the `radius`. Any ideas on how we should proceed here? At the moment I don't have a good name for this public API function that returns 1 in the case of ellipsoid and `radius(neigh)` in the case of ball.

juliohm (changes requested):

juliohm:
The issue boils down to the following:

For Euclidean distance and other distances, we have the call `inrange(metric, radius)` to collect all points inside the ball. However, for the ellipsoid case with the Mahalanobis distance, the call is `inrange(metric, 1)`, i.e. it assumes that the metric has been properly scaled to accomodate the semiaxes and rotations. I looked into Distances.jl and into NearestNeighbors.jl to figure out if there was a way to re-scale a metric so that its effect radius becomes 1 but couldn't find a way. That would be an elegant solution I guess.

If it is not possible to re-scale all metrics to a unit radius, we can think about a function in this public api that accesses this "one" in the case of ellipsoid and "radius" in the case of balls. I am starting to think that the previous `radius(ellips)` was enough, but I am not sure. It seems unnatural to define a single radius for something that is anisotropic.

rmcaixeta:
Yeah, it's more a naming issue.
I think `range()` would be a more appropriate name to call `one()` or `radius()`, but could be confusing as there is already a  `Base.range()`. Maybe `ballrange`?

juliohm:
Maybe it is more than just a naming issue. We are adopting two different
conventions for these balls where one is scaled to unit "range" and
another is not. I wonder if we can refactor the code to remove this
inconsistency somehow.

Em seg., 8 de fev. de 2021 às 16:51, Rafael Caixeta <
notifications@github.com> escreveu:

juliohm:
Perhaps we should concentrate on the `search` methods, meaning we could refactor the common parts of the code for different that are unrelated to the neighborhood types (e.g. treatment of the `mask` option) and then provide an intermediate function that dispatches on different neighborhood types like

```julia
searchinds(xₒ::AbstractVector, method::NeighborhoodSearch{O,N,T}) where {O,N<:BallNeighborhood,T} =
  inrange(method.tree, xₒ, radius(method.neigh))
searchinds(xₒ::AbstractVector, method::NeighborhoodSearch{O,N,T}) where {O,N<:EllipsoidNeighborhood,T} =
  inrange(method.tree, xₒ, one(eltype(xₒ)))
searchinds(xₒ::AbstractVector, method::NeighborhoodSearch) = # fallback to slow search method
```

Inside each of these `searchinds` function we could write a specific code for the neighborhood type. That will save us from having to think about a better name for `radius`in the Ellipsoid case or having to think about re-scaling features. What do you think? Is the proposal reasonable?

rmcaixeta:
Ok. I'll do that later today or tomorrow

rmcaixeta:
Made the changes for Ball and Ellipsoid, working nice
The generic fallback I didn't change yet because I'm not sure how to test it. Do you have an example of usage for it? Didn't find anything in the tests

juliohm:
That is awesome. I will review the PR before the end of the day.

Regarding the generic fallback, we should probably add a dummy neighborhood type and test it just to increase test coverage, probably in a future PR.

src/neighborhoods/ellipsoid.jl
juliohm: Did you have a chance to commit this suggestion @rmcaixeta ?

src/neighborsearch/neighborhood.jl
```diff
@@ -26,9 +26,16 @@ function NeighborhoodSearch(object::O, neigh::N) where {O,N}
...
     nothing
   end
+
   NeighborhoodSearch{O,N,typeof(tree)}(object, neigh, tree)
 end
 
+searchinds(xₒ::AbstractVector, method::NeighborhoodSearch{O,N,T}) where {O,N<:BallNeighborhood,T} =
+  inrange(method.tree, xₒ, radius(method.neigh))
+
+searchinds(xₒ::AbstractVector, method::NeighborhoodSearch{O,N,T}) where {O,N<:EllipsoidNeighborhood,T} =
```
juliohm, line 36:
```suggestion
searchinds(xₒ, method::NeighborhoodSearch{O,N,T}) where {O,N<:BallNeighborhood,T} =
  inrange(method.tree, xₒ, radius(method.neigh))

searchinds(xₒ, method::NeighborhoodSearch{O,N,T}) where {O,N<:EllipsoidNeighborhood,T} =
```

src/neighborsearch/neighborhood.jl
```diff
@@ -26,9 +26,16 @@ function NeighborhoodSearch(object::O, neigh::N) where {O,N}
...
+  inrange(method.tree, xₒ, one(eltype(xₒ)))
+
 # search method for any neighborhood
 function search(xₒ::AbstractVector, method::NeighborhoodSearch; mask=nothing)
```
juliohm, line 40: Perhaps we can also omit the type annotation of the first argument here too. Could you please double check?

juliohm (changes requested):
Just a few minor comments before we can merge it 👍🏽

src/neighborhoods/ellipsoid.jl
rmcaixeta: Haven't noted that one. Now it's adjusted

src/neighborsearch/neighborhood.jl
rmcaixeta: Done

src/neighborsearch/neighborhood.jl
rmcaixeta: Omitting this one leads to ambiguous issues during compiling. I maintained as it is
Question: when it's recommended to omit or not the type annotation? Is it just to avoid ambiguities or does it help on performance too?

juliohm: It doesn't affect performance, but it makes the function more generic in case we want to support other types of input in the future. Adding type annotation to function arguments without a good reason is a Julia anti-pattern. I did that myself in the past and now whenever I encounter an annotation in the codebase I try to clean it up. See the section "Over-constraining argument types" in this blog post: https://www.oxinabox.net/2020/04/19/Julia-Antipatterns.html

juliohm (approved):

juliohm:
The tests on Linux are failing due to an ongoing switch to ReferenceTests.jl and Julia v1.6.

Merging this one 👍🏽

PR #112, opened 2022-10-16
New groupby implementation

eliascarv:
This PR adds a new implementation for `groupby`.
Now `groupby` is a macro that supports grouping Mashes.Data objects in more than one column.
Column selection is now provided by the `ColSpec` interface.

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -1,5 +1,0 @@
-# ------------------------------------------------------------------
-# Licensed under the MIT License. See LICENSE in the project root.
-# ------------------------------------------------------------------
-
-@deprecate groupby(data::Data, var::Symbol) _groupby(data, var)
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -1,0 +1,5 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+@deprecate groupby(data::Data, var::Symbol) _groupby(data, var)
--- a/src/geoops/groupby.jl
+++ b/src/geoops/groupby.jl
@@ -2,44 +2,27 @@
 # Licensed under the MIT License. See LICENSE in the project root.
 # ------------------------------------------------------------------
 
-"""
-    groupby(data, var)
+macro groupby(data, cols...)
+  :(_groupby($(esc(data)), $(cols...)))
+end
 
-Partition spatial `data` into groups of constant value
-for spatial variable `var`.
+_groupby(data::Data, spec) = _groupby(data, colspec(spec))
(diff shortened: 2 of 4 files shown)


src/geoops/groupby.jl
```diff
@@ -3,43 +3,48 @@
...
-### Notes
+    @groupby(data, regex)
 
-Missing values are grouped into a separate group.
-"""
-function groupby(data::Data, var::Symbol)
-  vars = Tables.schema(values(data)).names
-  @assert var ∈ vars "invalid variable name"
+Partition spatial `data` into groups of constant value
+for columns that math with `regex`.
```
juliohm, line 16:
```suggestion
Partition geospatial `data` into groups of constant value
for columns that match with `regex`.
```

src/geoops/groupby.jl
```diff
@@ -3,43 +3,48 @@
 # ------------------------------------------------------------------
 
 """
-    groupby(data, var)
+    @groupby(data, col₁, col₂, ..., colₙ)
+    @groupby(data, [col₁, col₂, ..., colₙ])
+    @groupby(data, (col₁, col₂, ..., colₙ))
 
 Partition spatial `data` into groups of constant value
```
juliohm, line 10:
```suggestion
Partition geospatial `data` into groups of constant value
```

juliohm (changes requested):
Few suggestions.

src/geoops/groupby.jl
```diff
@@ -3,43 +3,48 @@
...
 
-  # retrieve value from each subset
-  vals = [vdata[s[1]] for s in indices(p)]
+  scolumns = [Tables.getcolumn(cols, nm) for nm in snames]
 
-  # save metadata
-  metadata = Dict(:values => vals)
+  srows = collect(zip(scolumns...))
+  urows = unique(srows)
+  inds  = map(row -> findall(Base.Fix2(===, row), srows), urows)
```
juliohm, line 46:
```suggestion
  inds  = map(row -> findall(===(row), srows), urows)
```

src/deprecations.jl
```diff
@@ -0,0 +1,5 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+@deprecate groupby(data::Data, var::Symbol) _groupby(data, var)
```
juliohm, line 5:
```suggestion
@deprecate groupby(data::Data, var::Symbol) @groupby(data, var)
```

juliohm (changes requested):
Amazing work @eliascarv. I think in this case we can just eliminate the old groupby entirely without deprecations. I will document the new function properly in the new release of the main package.

juliohm (approved):

PR #114, opened 2022-10-20
`@transform` macro

eliascarv:
This PR implements the `@transform` macro.
This macro takes as argument a `data` object of type `Meshes.Data` and a list of transformations and returns a new `Meshes.Data` object with `data` columns and new columns calculated using the transformations.

Examples:
```julia
x = [1, 1, 1, 1, 2, 2, 2, 2]
y = [1, 1, 2, 2, 3, 3, 4, 4]
table = (; x, y)
sdata = georef(table, rand(2, 8))

ndata = @transform(sdata, :z = :x + 2*:y)
values(ndata)
```
```julia
julia> values(ndata)
Table with 3 columns and 8 rows:
     x  y  z
   ┌─────────
 1 │ 1  1  3
 2 │ 1  1  3
 3 │ 1  2  5
 4 │ 1  2  5
 5 │ 2  3  8
 6 │ 2  3  8
 7 │ 2  4  10
 8 │ 2  4  10
```


src/geoops.jl
```diff
@@ -5,4 +5,5 @@
 include("geoops/filter.jl")
 include("geoops/groupby.jl")
 include("geoops/unique.jl")
-include("geoops/integrate.jl")
\ No newline at end of file
+include("geoops/integrate.jl")
+include("geoops/transform.jl")
```
juliohm, line 9: let`s move the include to after the groupby

src/geoops/transform.jl
```diff
@@ -0,0 +1,100 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    @transform(data, :newcol₁ = expr₁, :newcol₂ = expr₂, ..., :newcolₙ = exprₙ)
+
+Return a new data object with `data` columns and new columns `newcol₁`, `newcol₂`, ..., `newcolₙ`.
+The new columns are calculated using their respective expressions.
+In expressions the `data` columns are represented by `Symbols` and the operations use `broadcast` by default.
```
juliohm, line 10:
```suggestion
Return a new data object with `data` columns and new columns
`newcol₁`, `newcol₂`, ..., `newcolₙ` defined by expressions
`expr₁`, `expr₂`, ..., `exprₙ`. In each expression the `data`
columns are represented by symbols and the functions
use `broadcast` by default.
```

src/geoops/transform.jl
```diff
@@ -0,0 +1,100 @@
...
+
+# Examples
+
+```julia
+@transform(data, :z = :x + 2*:y)
+@transform(data, :w = :x^2 - :y^2)
+@transform(data, :sinx = sin(:x), :cosy = cos(:y))
+```
+"""
+macro transform(data::Symbol, exprs...)
```
juliohm, line 20: This argument name is confusing, it is not geospatial data, right?

src/geoops/transform.jl
```diff
@@ -0,0 +1,100 @@
...
+  splits = map(expr -> _split(data, expr), exprs)
+  colnames = first.(splits)
+  colexprs = last.(splits)
+  texpr = :(GeoStatsBase._transform($data, [$(colnames...)], [$(colexprs...)]))
+  esc(texpr)
+end
+
+function _transform(data::D, tnames, tcolumns) where {D<:Data}
```
juliohm, line 28:
```suggestion
function _transform(data::Data, tnames, tcolumns)
```

src/geoops/transform.jl
```diff
@@ -0,0 +1,100 @@
...
+
+  @assert isdisjoint(tnames, names) "Invalid column names"
+
+  newnames   = [names; tnames]
+  newcolumns = [columns; tcolumns]
+
+  𝒯 = (; zip(newnames, newcolumns)...)
+  newtable = 𝒯 |> Tables.materializer(table)
+
+  newdata = Dict(paramdim(dom) => newtable)
```
juliohm, line 44: we only use the term "data" for geospatial data. Here we are generating a dictionary of tables. Try to follow the naming convention in other parts of the code.

src/geoops/transform.jl
```diff
@@ -0,0 +1,100 @@
...
+  constructor(D)(dom, newdata)
+end
+
+# macro utils
+function _split(data::Symbol, expr::Expr)
```
juliohm, line 49: the name data is used for geospatial data

juliohm (changes requested):
Very nice to see this getting shape 💯

src/geoops.jl
```diff
@@ -4,5 +4,6 @@
 
 include("geoops/filter.jl")
 include("geoops/groupby.jl")
+include("geoops/transform.jl")
```
juliohm, line 7: let's move these two includes to the end of the list after integrate

src/geoops/transform.jl
```diff
@@ -0,0 +1,108 @@
...
+@transform(data, :z = :x + 2*:y)
+@transform(data, :w = :x^2 - :y^2)
+@transform(data, :sinx = sin(:x), :cosy = cos(:y))
+```
+"""
+macro transform(data::Symbol, exprs...)
+  splits = map(expr -> _split(data, expr), exprs)
+  colnames = first.(splits)
+  colexprs = last.(splits)
+  texpr = :(GeoStatsBase._transform($data, [$(colnames...)], [$(colexprs...)]))
```
juliohm, line 27: Do we need the prefix GeoStatsBase here? The _split is used without prefix

juliohm (approved):

juliohm:
Beautiful! 💯

PR #115, opened 2022-11-04
Macro Utils

eliascarv:


src/geoops.jl
```diff
@@ -2,6 +2,7 @@
 # Licensed under the MIT License. See LICENSE in the project root.
 # ------------------------------------------------------------------
 
+include("geoops/macroutils.jl")
```
juliohm, line 5: let's rename to geoops/macros.jl

juliohm (changes requested):
Let's just migrate the macros to a separate file in this PR and come back to it after we fix the other issues we have identified in other modules.

juliohm (approved):

PR #135, opened 2023-05-17
Add `UniqueCoords` transform

eliascarv:
This PR removes the `uniquecoors` function and replaces it with the `UniqueCoords` transform.


src/transforms/uniquecoords.jl
```diff
@@ -52,39 +51,30 @@ function apply(transform::UniqueCoords, data::Data)
...
   pts = [centroid(dom, i) for i in 1:nelements(dom)]
   X = reduce(hcat, coordinates.(pts))
-  ginds = _uniqueinds(X, 2)
-  groups = Dict(ind => Int[] for ind in unique(ginds))
-  for (i, ind) in enumerate(ginds)
-    push!(groups[ind], i)
-  end
+  uinds = _uniqueinds(X, 2)
+  ginds = unique(uinds)
+  groups = [findall(==(ind), uinds) for ind in ginds]
```
eliascarv, line 56: I couldn't think of a more efficient way to collect group indices. Do you have any suggestions @juliohm?

test/transforms.jl
```diff
@@ -145,6 +145,100 @@
     end
   end
 
+  @testset "UniqueCoords" begin
+    X = [i * j for i in 1:2, j in 1:100]
```
juliohm, line 149: We tested with 1_000_000 to make sure that the implementation scales. If we remove the Base implementation of uniqueinds for example the algorithm takes forever. Can you please use 1_000_000 again?

src/transforms/uniquecoords.jl
eliascarv: Resolved!

juliohm (approved):

PR #142, opened 2023-06-13
[WIP] Add `describe` function

eliascarv:


juliohm (changes requested):
Awesome WIP 💯

src/geoops/describe.jl
```diff
@@ -0,0 +1,56 @@
...
+describe(data::Data, cols::T...; kwargs...) where {T<:Col} = describe(data, colspec(cols); kwargs...)
+
+function _describe(data::Data, colspec::ColSpec, funs::Vector{Function})
+  table = values(data)
+  cols = Tables.columns(table)
+  names = Tables.columnnames(cols)
+  snames = choose(colspec, names)
+
+  dnames = [:variable; nameof.(funs)]
+  dcolumns = Vector[snames]
```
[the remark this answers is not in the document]
eliascarv: `dcolumns` is a vector of vectors

src/geoops/describe.jl
```diff
@@ -0,0 +1,56 @@
...
+  for fun in funs
+    column = map(snames) do name
+      try
+        x = Tables.getcolumn(cols, name)
+        fun(x)
+      catch
+        nothing
+      end
+    end
+    push!(pairs, nameof(fun) => column)
```
juliohm, line 52:
```suggestion
    funvalue = map(snames) do name
      try
        x = Tables.getcolumn(cols, name)
        fun(x)
      catch
        nothing
      end
    end
    push!(pairs, nameof(fun) => funvalue)
```

juliohm (changes requested):
Only missing tests now 👍🏽

Project.toml
```diff
@@ -5,6 +5,7 @@ version = "0.31.4"
 
 [deps]
 Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
+DataAPI = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
```
juliohm, line 8: We can get the `describe` name directly from StatsBase.jl, which is a better dependency than DataAPI.jl in this case.

src/geoops/describe.jl
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  TypedTables.Table(; pairs...)
```
juliohm, line 55: Do we need to prefix with TypedTables? If not, please remove the prefix.

eliascarv: Yes, because both TypedTables and ScientificTypes exports the name `Table`

juliohm (approved):
Very nice. Just waiting for tests.

